PR #397, opened 2026-04-24
370 auth prepare base authentication structure in account service

### LINKED ISSUE #370: [AUTH] Prepare base authentication structure in account-service
## Goal

Create the minimal backend foundation required to support authentication.

## Scope

- Define a basic package/class structure for authentication  
- Add initial configuration related to `security/OAuth`  
- Avoid replicating the full structure of other modules/services at this stage  

## Acceptance Criteria

- A clear base structure for authentication exists  
- The account-service starts correctly after the changes  

## Minimal Validation

- The service runs successfully without errors after introducing the base structure

EstherBlacksmith:
Base issue: #370

Goal
---
Prepare the base structure to allow external authentication by GitHub.

What has been done
---

- Added the necessary Gradle dependencies
- Added the `SecurityConfig.java` file with the basic routes
- Added Oauth2 in the `application.yaml ` file
- Created some basic tests to assure the correct running


backend/account-service/src/main/java/com/ita/challenges/account/infraestructure/config/SecurityConfig.java
```diff
@@ -0,0 +1,26 @@
+package com.ita.challenges.account.infraestructure.config;
```
eariasvalor: Little eye! There is a typo here: infraestructure > infrastructure

backend/account-service/src/test/java/com/ita/challenges/account/infraestructure/config/SecurityConfigTest.java
```diff
@@ -0,0 +1,54 @@
+package com.ita.challenges.account.infraestructure.config;
```
eariasvalor: typo as above

backend/account-service/src/test/java/com/ita/challenges/account/infrastructure/config/SecurityConfigTest.java
```diff
@@ -0,0 +1,54 @@
...
+    @Autowired  
+    private SecurityFilterChain securityFilterChain;  
+  
+    @Test
+    @DisplayName("Should assure the SecurityFilterChain is loaded")
+    void contextLoads() {  
+        assertThat(securityFilterChain).isNotNull();  
+    }  
+  
+    @Test
```
eariasvalor, line 33: I think that, according to the name of the test, if it only expects "isNotFound", ie 404, it only proves that the endpoint doesn't exist. Maybe it should be **not** expecting a 401 or a 302 to be consistent with its purpose

backend/account-service/src/test/java/com/ita/challenges/account/infraestructure/config/SecurityConfigTest.java
```diff
@@ -0,0 +1,54 @@
...
+import org.springframework.boot.test.autoconfigure.web.servlet.AutoConfigureMockMvc;  
+import org.springframework.boot.test.context.SpringBootTest;  
+import org.springframework.security.web.SecurityFilterChain;  
+import org.springframework.test.web.servlet.MockMvc;  
+  
+import static org.springframework.test.web.servlet.request.MockMvcRequestBuilders.get;  
+import static org.springframework.test.web.servlet.result.MockMvcResultMatchers.*;  
+import static org.assertj.core.api.Assertions.assertThat;  
+  
+@SpringBootTest(webEnvironment = SpringBootTest.WebEnvironment.RANDOM_PORT)
```
eariasvalor: In my opinion, there is no need to load the entire context for a config test, so I recommend using @WebMvcTest with @Import(SecurityConfig.class, it would be lighter to run

eariasvalor (changes requested):
Great Job, Esther! Only a couple of comments about a typo (infraestructure) and test performance.

backend/account-service/src/main/java/com/ita/challenges/account/infraestructure/config/SecurityConfig.java
EstherBlacksmith: 😨 Shame on me. You're right.

backend/account-service/src/test/java/com/ita/challenges/account/infraestructure/config/SecurityConfigTest.java
EstherBlacksmith: 😨

backend/account-service/src/test/java/com/ita/challenges/account/infrastructure/config/SecurityConfigTest.java
EstherBlacksmith: Not exactly. We're using a pattern to match every possible route (/api/account/**), and in the call of the test we're using one example to test if a random request would be allowed. The way to know that it is allowed but catching the error if the route doesn't exist is with 404. 
401 ('Not Authorized') or 302 ('Founded' but temporarily unable) are not statuses that you can check in a test without a specific request. 
In my opinion, for the purpose of the test (not finding a route  but allowing the request) 'Not found' is a correct option.

backend/account-service/src/test/java/com/ita/challenges/account/infraestructure/config/SecurityConfigTest.java
EstherBlacksmith: You're right, with these annotations the test will be lighter ✨

backend/account-service/src/test/java/com/ita/challenges/account/infrastructure/config/SecurityConfigTest.java
eariasvalor: Ok, I take that back!

eariasvalor (approved):
Thanks for the clarification, I wasn't even sure about it, just wondered. Everything else looks good to me now.

andyrouge81 (approved):
Nice update, Esther

The corrections look good to me, especially the package naming fix and the lighter test setup using `@WebMvcTest`. The changes are clear and aligned with the goal of preparing the base `OAuth2` structure.

Everything looks good from my side.

AlbertMedina (approved):
Good job Esther! All clear now after the updates. LGTM!

backend/account-service/src/main/java/com/ita/challenges/account/infrastructure/config/SecurityConfig.java
```diff
@@ -0,0 +1,26 @@
...
+
+@Configuration
+@EnableWebSecurity
+public class SecurityConfig {
+
+    @Bean
+    public SecurityFilterChain filterChain(HttpSecurity http) throws Exception {
+        http
+                .authorizeHttpRequests(auth -> auth
+                        .requestMatchers("/api/account/**").permitAll()
```
yvangabrieli: I noticed /api/account/** was fully public. I’ve suggested restricting it to /api/account/auth/** so only auth endpoints stay open. This secures the rest of the account routes while allowing the oauth2Login redirect to work perfectly.

yvangabrieli (changes requested):

backend/account-service/src/main/java/com/ita/challenges/account/infrastructure/config/SecurityConfig.java
eariasvalor: Hi Yvan, I think that this s could be nice to do in the future if it's needed, but now, given that this is the base for the authentication and that the scope of the PR is minimalist, it actually should be public for the auth flow to work, shouldn't it?

eariasvalor: Sorry, I take it back, now I see you are right, */auth is absolutely necessary here.

EstherBlacksmith: Guys, you’re both right. On the one hand, security is very weak and on the other hand we don’t have the definitive routes. I modify what Yvan says, to ensure a little more security. But keep in mind that this file will be alive until we have all the clear routes

yvangabrieli (approved):
Approve. Great job

yvangabrieli (approved):
Correct modification on the test. I would just suggest naming the method (shouldPermitAllRequestsToAccountApi) accordingly (shouldPermitAllRequestsToAuth)

andyrouge81 (approved):
Re-approving after the latest updates.

There are already multiple approvals, but since I participated in the previous review, I’m approving again to reflect the latest changes.

GJ!

backend/account-service/src/test/java/com/ita/challenges/account/infrastructure/config/SecurityConfigTest.java
kenan-it-academy: I disagree with both of you 🤣. If the test is about the existence of the endpoint, test that it returns something that tells you it exists. NotFound does not do this.
I would say that the simplest way is to find what an endpoint returns when it is NOT permitted and check against that. Intuitively, as the language of the test says "should permit" I would assume that the intention would be to check that you DON'T get a 403 - Forbidden result.
Please think about how to test it thoroughly and reevaluate the test itself if necessary.

kenan-it-academy (changes requested):

backend/account-service/src/test/java/com/ita/challenges/account/infrastructure/config/SecurityConfigTest.java
EstherBlacksmith: I made some changes (added a controller test) to allow a better tests and provide explicit verification that the the security configuration allows all requests. The test now confirms requests pass through security filter chain and are handled by controller (rather than just not being blocked).

andyrouge81 (commented):
I know you were working on this test yesterday, but I’m not sure if you aligned it with the tech lead in the end. Right now it looks a bit inconsistent with the controller, since the test calls `/api/account/auth/test` but the controller exposes `/api/account/auth/test-ok`.

I’m not sure if the goal of the test is to keep expecting a 404 or if you intended it to return a 200 OK.

backend/account-service/src/test/java/com/ita/challenges/account/infrastructure/config/SecurityConfigTest.java
eariasvalor: Hi Esther, I see where you are going with the new controller.
Personally, I don't think it's necessary to expose a new endpoint only to check if another endpoint works. I would stick to Kenan's proposal to check in this test that the endpoint doesn't return a 403 Forbidden, as it would mean that the user has indeed authenticated and therefore the endpoint exists and does its work.

backend/account-service/src/test/java/com/ita/challenges/account/infrastructure/config/TestAuthController.java
```diff
@@ -0,0 +1,13 @@
+package com.ita.challenges.account.infrastructure.config;
+
+import org.springframework.web.bind.annotation.GetMapping;
+import org.springframework.web.bind.annotation.RestController;
+
+@RestController
+public class TestAuthController {
```
eariasvalor: If you decide to change the other test as per Kenan's proposal, this entire controller would have to be removed.

eariasvalor (changes requested):
Hi Esther! Please give my recommendations a think, specifically about the need of a new endpoint.

backend/account-service/src/test/java/com/ita/challenges/account/infrastructure/config/SecurityConfigTest.java
EstherBlacksmith: You're right, isn't the best practice. I rethought the test itsefl and follow your both previous suggestions, made another change and deleted the controller test file. 
Thank you for the comments and the feedback! ❤️

eariasvalor (approved):
Hi Esther! Everything looks OK now! Approved from my side.

kenan-it-academy (approved):

PR #405, opened 2026-04-25
feat(auth): Implement GitHub Oauth2 callback and retreave user data

### LINKED ISSUE #372: [AUTH] Handle GitHub OAuth2 callback and retrieve user data
## Goal

Handle the GitHub OAuth2 callback and retrieve the minimum required user data.

## How the GitHub callback works

- GitHub does not directly send user data to the callback  
- After authentication, GitHub redirects to the configured callback URL with query parameters  

In the happy path, the backend typically receives:

- `code`: temporary authorization code  
- `state`: value used to validate the request  

Example:

`GET /login/oauth/callback?code=abc123&state=xyz789``

- The `code` is not the final token or user data  
- The backend must exchange this `code` for an access token from GitHub  
- Once the access token is obtained, the backend calls GitHub’s API to retrieve user data (e.g. username)  

In case of error, GitHub may return:

- `error`  
- `error_description`  
- `error_uri`  
- `state`  

## Flow Summary

- Frontend initiates login via backend (through Nginx)  
- Backend redirects to GitHub OAuth login  
- GitHub authenticates the user  
- GitHub redirects back to the backend callback with `code` and `state`  
- Backend exchanges the `code` for an access token  
- Backend uses the token to fetch user data from GitHub  
- Backend returns minimal user information (e.g. username) to the frontend  

## Scope

- Handle the OAuth2 callback  
- Retrieve at least the GitHub username  
- Optionally retrieve `avatarUrl`  

## Acceptance Criteria

- After successful login, the backend retrieves the minimal GitHub user identity  
- The flow works without requiring a database or JWT  
- The implementation does not depend on challenge-service or shared user identity  

## Minimal Validation

- Verify that the callback successfully retrieves at least the username  
- If tests are included, prioritize a simple test validating minimal mapping/response

yvangabrieli:
Goal
Handle the GitHub OAuth2 callback and retrieve the minimum required user data.

Implementation
Updates the success endpoint to return the real GitHub username and
avatar URL after login instead of a hardcoded message.

Captura de pantalla 2026-04-25 a las 2 08 02

Endpoint response
GET /api/account/success
{
  "username": "github-username",
  "avatarUrl": "https://avatars.githubusercontent.com/..."
}


andyrouge81 (changes requested):
HI Yvan,

I noticed a small issue in the avatar field: the attribute name should be `avatar_url`, but it’s currently written as `avaRtar_url`, so it may not return the expected value.

Also, in the test:
`@DisplayName("Should permit all requests to /api/account/** endpoints")`
this doesn’t fully match the current security config (/api/account/auth/**), so it might be worth aligning it to avoid confusion.

Apart from that, everything looks good.

yvangabrieli:
PR will be close for generating confusion on the merging process.

PR #407, opened 2026-04-25
feat(challenge-service): delete function to remove challenge (#368)

### LINKED ISSUE #368: [CRUD] [FE] Enable the delete function to remove a challenge (#343)
## Description

**As a** user, **when** I click the delete button on the deletion screen, **I want** the system to execute the removal command, **so that** the selected challenge is successfully deleted.

## Context

This task focuses on the internal logic required to process a deletion request. It follows a **"happy path"** execution, meaning we assume the operation will always be successful without needing to handle errors or invalid IDs at this stage.

The implementation connects the user interface to the backend services through two layers:
1. **ChallengeService**: You must implement the logic in the `delete(id)` method. This method acts as a coordinator that receives the request and passes it to the API layer.
2. **ChallengeAPIService**: The `delete(id)` method in this service handles the communication with the external data source. For this task, it should be configured to always return a successful operation response.

The goal is to ensure that the command triggered by the user travels correctly through the service layers to simulate a complete deletion.

## Out of scope

* **Confirmation Dialogs:** Adding "Are you sure?" pop-ups or secondary warnings.
* **Error Handling:** Managing cases where the challenge ID is not found or the server is down.
* **UI Design:** Modifying the layout of the deletion screen or the button styles.
* **Permissions:** Verifying if the current user has the authority to delete the item.

## Acceptance Criteria

* The `ChallengeService.delete(id)` method is correctly implemented.
* the `ChallengeService` successfully calls the `ChallengeAPIService.delete(id)` method.
* The `ChallengeAPIService` always returns a "success" response as per the happy path requirement.
* The specific challenge **ID** is passed correctly from the service layer to the API layer.

soyjuandelgado:
Base issue: **#368**

## Goal

Connect `ChallengeService.delete()` with `ChallengeAPIService.delete()`.

## What has been done

* **Logic:** Added a call from `delete()` in `ChallengeService` to `ChallengeAPIService`.
* **Testing:** Implemented unit tests for the new logic.

## What has NOT been done

* **API connection:** It requires previous configuration and the endpoint base URL.

### DIFF
--- a/frontend/src/app/features/challenges/services/challenge.service.spec.ts
+++ b/frontend/src/app/features/challenges/services/challenge.service.spec.ts
@@ -1,16 +1,40 @@
 import { TestBed } from '@angular/core/testing';
+import { of } from 'rxjs';
 
 import { ChallengeService } from './challenge.service';
+import { ChallengeApiService } from '../data-access/challenge-api.service';
 
 describe('ChallengeService', () => {
   let service: ChallengeService;
+  let mockChallengeApiService: Partial<ChallengeApiService>;
 
   beforeEach(() => {
-    TestBed.configureTestingModule({});
+    mockChallengeApiService = {
+      delete: (id: string) => of(undefined)
+    };
+
+    TestBed.configureTestingModule({
+      providers: [
+        { provide: ChallengeApiService, useValue: mockChallengeApiService }
+      ]
+    });
     service = TestBed.inject(ChallengeService);
   });
 
   it('should be created', () => {
     expect(service).toBeTruthy();
+  });
+
+  describe('delete', () => {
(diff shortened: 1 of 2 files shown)


JordiMiravet (approved):
LGTM

andyrouge81 (approved):
Nice work Juan,

It’s well structured and the implementation is clear. The tests also cover the happy path nicely.

As a small improvement, you could consider adding a test for the error case as well, so the behavior is fully covered. That would make it even more complete.

GJ.

nereame96 (approved):
Looks okey, good job!

kenan-it-academy (approved):

PR #409, opened 2026-04-25
feat: add findAllChallenges method signature to ChallengeRepository port

JordiCasas87:
This PR covers the first step for the `findAll` flow by defining the output port contract in the catalog domain.

- add `findAll` to `ChallengeRepository`
- add a provisional override in `InMemoryChallengeRepository`
- leave the real adapter logic for the next sub-issue

## Notes
- no real `findAll` implementation yet, no controller changes included
- compile-safe provisional stub added in the in-memory adapter

Closes #331

PR #410, opened 2026-04-25
332 crud findall retrieve all existing challenges from storage

### LINKED ISSUE #332: [CRUD] [FindAll] Retrieve all existing challenges from storage
## Description

Implement the `findAllChallenges` method in the infrastructure adapter, fulfilling the contract defined in the output port.

## Acceptance Criteria

- [ ] Adapter implements the output port method.
- [ ] Correctly returns all existing challenges.
- [ ] Unit tests passing (TDD).

eariasvalor:
This PR adds findAll to the ChallengeRepository port and implements it in InMemoryChallengeRepository returning all stored challenges.

Only the empty list test is included, the test for actual retrieval depends on save, which is being handled in a separate issue and will be added once merged into develop.


JordiCasas87 (approved):
Looks good. findAll is correctly implemented and the empty case is covered. Including the interface change here makes sense given the pending contract PR. Nice work.

andyrouge81 (approved):
Nice work Esther,

The implementation looks clear and the empty case is well covered.

One small thing I noticed: the related contract PR defines `findAllChallenges()` in #409, while here the method is named `findAll()`. Since both represent the same operation, it might be worth aligning the naming to avoid duplication — although I assume this was already discussed when splitting the tasks.

Apart from that, everything looks good from my side.

Andrescalvo22 (changes requested):
Overall, this PR is well implemented and clearly focused on the findAll use case.
The repository implementation using ConcurrentHashMap is simple and appropriate for an in-memory adapter, and the test correctly validates the empty state!

Small point to consider:

The test is currently located in:

catalog.infrastructure.adapter.out.persistence

while the repository is in:

catalog.infrastructure.adapter.web.out.persistence

It might be worth aligning both packages to keep consistency and avoid potential visibility issues (especially if we rely on package-private access later on).

eariasvalor:
Hi guys, both of you are right. I'm changing the path of the test. As for the naming of the method to find all challenges, if we follow Spring conventions it should be findAll(). @JordiCasas87, what do you think? Would you agree on changing the contract method? findAllChallenges in ChallengeRepository is also a bit redundant.

JordiCasas87:
@eariasvalor I agree, findAll() fits better with Spring conventions and keeps things cleaner. 👍

Andrescalvo22 (approved):
Now is perfectly alligned, great Job!

AlbertMedina (approved):
Nice work! The fixes and refactor look great and everything is clear now. LGTM!

eariasvalor:
Ok, I pushed the wrong commits to this branch, then freaked out, then reverted them, this is now fixed. I'm cherry picking them and taking them to the right branch. Sorry about the mess.

backend/challenge-service/src/main/java/com/itachallenges/challengeservice/catalog/domain/port/out/ChallengeRepository.java
```diff
@@ -1,4 +1,10 @@
 package com.itachallenges.challengeservice.catalog.domain.port.out;
 
+import com.itachallenges.challengeservice.catalog.domain.model.Challenge;
+
+import java.util.List;
+
 public interface ChallengeRepository {
+    List<Challenge> findAll();
+    void save(Challenge challenge);
```
kenan-it-academy: What is the `save` related doing here?

kenan-it-academy (changes requested):
Remove out-of-scope save implementation from this task. Also I find it funny that this task does have the test when the others don't 🤣

backend/challenge-service/src/main/java/com/itachallenges/challengeservice/catalog/domain/port/out/ChallengeRepository.java
eariasvalor: errr let's say I tried different things before this one and this is only a vestige of it. Removing it.

kenan-it-academy (approved):

PR #413, opened 2026-04-25
Feature/implement update in infraestructure

Andrescalvo22:
## Description

Implemented the update method in the infrastructure adapter to fulfill the contract defined in the domain output port.
Also added a restore factory method in the domain model to support reconstruction of existing challenges.

---

## Changes

- InMemoryChallengeRepository: implemented update method using in-memory storage.
- Added unit test should_update_existing_challenge to validate update behavior.
- Challenge: added restore method to recreate a challenge with an existing ID.

---

## Implementation details

- The repository uses a ConcurrentHashMap to simulate persistence.
The update method checks whether the challenge exists before updating it and replaces the existing entry with the new one.
If the challenge does not exist, a RuntimeException is thrown.

- Since the update operation requires preserving the original ID, a restore factory method was introduced in the domain model. This allows recreating a Challenge instance with an existing identifier, which is necessary for both update logic and testing.

- The test prepares the initial state by inserting a challenge into the in-memory storage before performing the update, ensuring the behavior matches real-world expectations.

- The use of the restore method and direct access to the in-memory storage in the test are temporary decisions due to the current task split.
Once a proper save method is implemented and merged, this can be refactored to rely on the repository's public API instead.


andyrouge81 (changes requested):
Just a couple of small suggestions:

* It might be worth making the in-memory `storage` field `private final` to keep encapsulation tighter and avoid exposing internal state.
* In the test, instead of accessing `repository.storage` directly, you could consider preparing the initial state through a repository method (e.g. `save()` if available), so the test relies on the public behavior rather than internal implementation details.

The update implementation looks solid and the test setup is clear and easy to follow. GJ

Andrescalvo22:
@andyrouge81 Thanks for the suggestions!

I agree that keeping the storage private and relying on a public method like save() would be a better approach in terms of encapsulation and test design.

In this case, since save() is not implemented yet, I accessed the in-memory storage directly and removed the private modifier to avoid issues in the tests.

Once save() is implemented and merged, we can refactor this to rely on the public API instead 👍

andyrouge81 (approved):
Thanks for the explanation, really helpful

Makes total sense to defer the encapsulation in this case and keep things simple until a proper `save()` (or similar public method) is in place.

Agree that it’s better to revisit this in a later stage once the repository API is more complete, so the tests can rely fully on the public behavior instead of internal state.

Appreciate the clarification!

JordiCasas87 (approved):
Hello! Looks good to me!

Small suggestion: if the use of the restore method and direct storage access in the test are temporary decisions due to the current task split, it could be helpful to briefly mention it in the PR description to provide more context for future readers.

AlbertMedina (approved):
Great job on the PR and handling the feedback. I agree with what's been said; now the details in the description add the necessary context and explain why this temporary solution with `storage` and `restore` was used. Once the `save` method is ready, we can refactor this to keep the internal state private again. Other than that, everything looks good to me!

backend/challenge-service/src/test/java/com/itachallenges/challengeservice/catalog/infrastructure/adapter/web/out/persistence/InMemoryChallengeRepositoryTest.java
```diff
@@ -0,0 +1,34 @@
...
+import static org.assertj.core.api.Assertions.assertThat;
+
+class InMemoryChallengeRepositoryTest {
+
+    private final InMemoryChallengeRepository repository = new InMemoryChallengeRepository();
+
+    @Test
+    void should_update_existing_challenge() {
+
+        // given
```
kenan-it-academy: Remove comments, they are unneeded

kenan-it-academy (changes requested):
Remove superfluous comments from code

Andrescalvo22:
@kenan-it-academy Comments Removed, Thanks for the observation!

kenan-it-academy (approved):

PR #419, opened 2026-04-26
feat(auth): Expose authenticated user for frontend via /me endpoint

### LINKED ISSUE #373: [AUTH] Define how to propagate authenticated user to frontend
## Goal

Define and implement a simple mechanism for the frontend to retrieve the authenticated user's username after a successful GitHub login.

## Acceptance Criteria
- A single clear mechanism is selected (e.g., redirect with data or /me endpoint)
- The frontend can reliably retrieve the authenticated user's username
- The data format and access method are clearly defined
- The mechanism works without requiring complex persistence or additional authentication layers

## Minimal Validation

After completing the GitHub login flow, the frontend successfully obtains and displays the authenticated user's username using the selected mechanism

yvangabrieli:
## Summary

Expose authenticated username to the frontend via /api/account/auth/me.

## Changes

- Added AuthUserDto
- Implemented AuthController with /me endpoint
- Added unit test using mocked Authentication

## Goal

Provide a simple way for the frontend to retrieve the authenticated username after login.

## Testing

Unit test validates controller logic in isolation from Spring Security

## Notes

Actual authentication flow depends on Spring Security configuration in the main branch.


andyrouge81 (changes requested):
The endpoint also correctly follows the established `/api/account/*` prefix, great.

One thing I noticed is that the current DTO only returns the 'username' and does not include `avatarUrl`. This might be perfectly fine depending on what has been agreed with the frontend, but it could be worth double-checking in case additional fields are expected.

Apart from that, there are just a few minor typographical and style details (like spacing and naming conventions). 
GJ!

yvangabrieli:
Thanks for the review!
Regarding avatarUrl — you're right, it is included in the DTO and returned by the endpoint. It might not have shown up clearly in the diff, but both username and avatarUrl are there.

andyrouge81 (changes requested):
Nice update, Yvan. The endpoint now looks much better aligned with the /`api/account/*` prefix, and returning both `username` and `avatarUrl`makes the frontend contract clearer.

I only noticed a couple of small things: there seems to be a typo in the package name, `infrastucture` instead of `infrastructure`. Also, in `build.gradle`, the `testRuntimeOnly` import appears in two different lines, one before the Spring Security test dependency and another after it, so it might be duplicated.

Also, the unauthenticated test expects a `302`, which makes sense if the expected Spring Security behaviour is redirecting to OAuth2 login, but maybe it is worth double-checking if that is the intended response for this API endpoint.

looks much better aligned now. GJ!

yvangabrieli:
Thanks for the review!

The `302` is actually intentional — since `/me` is the landing point after GitHub's redirect, Spring Security's OAuth2 flow is browser-based by design. A `401` would make more sense in a stateless JWT setup, but here the redirect is the correct behaviour per our `SecurityConfig`. The frontend always reaches `/me` through the login flow, never directly.

Good catch on the `testRuntimeOnly` duplicate and the `infrastucture` typo — both fixed! 👍

andyrouge81 (approved):
Nice update, Yvan. I see you’ve fixed the small typos and added `avatarUrl`, much clearer now 

About the Gradle wrapper downgrade, I assume it was due to some version incompatibility, but it might be worth double-checking since it could bring issues later on.

Looks good from my side. GJ!

yvangabrieli:
You are assuming correctly. I wrapper downgrade has been suggested from the SonarQube test.

backend/account-service/src/main/java/com/ita/challenges/account/infrastructure/adapter/web/in/AuthController.java
```diff
@@ -0,0 +1,23 @@
...
+import org.springframework.web.bind.annotation.RestController;
+
+@RestController
+@RequestMapping("/api/account")
+public class AuthController {
+
+    @GetMapping("/me")
+    public ResponseEntity<AuthUserDto> me(@AuthenticationPrincipal OAuth2User user) {
+        AuthUserDto dto = new AuthUserDto(
+                user.getAttribute("login"),     // GitHub username
```
eariasvalor: Please remove comments from the files

eariasvalor: if getAttribute("login") and getAttribute("avatar_url") return null, the dto will be built with nulls and there is a potential null pointer exception in /me, please add a fallback just in case

backend/account-service/build.gradle
```diff
@@ -11,9 +11,10 @@ dependencies {
 	implementation 'org.springframework.boot:spring-boot-starter-validation'
 	implementation 'org.springframework.boot:spring-boot-starter-security'
 	implementation 'org.springframework.boot:spring-boot-starter-actuator'
-
+    implementation 'org.springframework.boot:spring-boot-starter-oauth2-client'
 	testImplementation 'org.springframework.boot:spring-boot-starter-test'
-	testRuntimeOnly 'org.junit.platform:junit-platform-launcher'
+    testImplementation 'org.springframework.security:spring-security-test' // 👈 add this
```
eariasvalor: Please remove comments

eariasvalor (changes requested):
Hey! 
-I think downgrading the Gradle wrapper is a bit unusual and I find it strange that Sonar asked for that. Would it be possible to explore an alternative solution?
-Please check my comments in the file
-Please ensure no //comments are left in the files

JordiCasas87 (changes requested):
Nice update overall, everything looks much more aligned now 👍

I’ve noticed a small thing that would be good to adjust before merging:
* There are some inline comments with icons in `build.gradle`, and also small leftover comments in other files (e.g. AuthController). It would be good to remove them to keep the code clean.

Apart from that, looks good to me 👍

jrffruit518-enrich:
> Nice update overall, everything looks much more aligned now 👍
> 
> I’ve noticed a small thing that would be good to adjust before merging:
> 
> * There are some inline comments with icons in `build.gradle`, and also small leftover comments in other files (e.g. AuthController). It would be good to remove them to keep the code clean.
> 
> Apart from that, looks good to me 👍

I agree with Andy's suggestion. Removing the icons in build.gradle and the redundant comments in the Controller will make the code look more professional and cleaner.

AlbertMedina (commented):
I'm with the others on cleaning up the comments. Aside from that, all looks good!

yvangabrieli:
Comments have been clean up thank you for the suggestions. I appreciate.

JordiCasas87 (approved):
Thanks for addressing the comments!

eariasvalor (changes requested):
Thanks for removing the comments, before approving the PR I think we have to ensure that the needs in this PR are enough to downgrade the Gradle wrapper.

backend/gradle/wrapper/gradle-wrapper.properties
```diff
@@ -1,6 +1,6 @@
 distributionBase=GRADLE_USER_HOME
 distributionPath=wrapper/dists
-distributionUrl=https\://services.gradle.org/distributions/gradle-8.14.4-bin.zip
+distributionUrl=https\://services.gradle.org/distributions/gradle-8.11.1-bin.zip
```
EstherBlacksmith: Hello there! Can you explain the reason for the version modified in this line? Maybe this type of change affects more than we can see in this moment. Also, a change to downgrade the version can make the app more sensitive to vulnerabilities.

yvangabrieli:
Restore the Gradle wrapper to 8.14.4. Thank you Esther GJ.

backend/gradle/wrapper/gradle-wrapper.properties
yvangabrieli: We aligned the Gradle wrapper to 8.11.1 to keep all environments consistent (local development, CI, Docker, and future deployment setups).
This project depends on having very predictable builds across multiple services (gateway, backend services, and frontend), so stability is more important than using the newest Gradle version.
We appreciate the double-check on this, as it’s a good point to discuss and could lead to issues later if versions are not aligned across environments.
Gradle updates can sometimes cause small but important differences in how builds run, especially with tools like Spring Boot and SonarQube. In a multi-service setup like ours, even small differences can create inconsistencies between local builds and CI.
For now, staying on a version that is already well aligned with the current pipeline helps ensure everything works the same everywhere.
Upgrading Gradle can be revisited later as a separate improvement once the system is fully stable.

eariasvalor (approved):
Well done!

EstherBlacksmith (approved):

PR #430, opened 2026-04-27
feat(challenge-api-service): get all challenges from the server

### LINKED ISSUE #425: [CRUD] [FE] Get all challenges from the server
## Description

**As a** user, **when** I call the API, **I want** to retrieve all challenges from the server.

## Objective

This PR implements a **"happy path"** for fetching challenges through the API.

The goal is to correctly establish a connection to the `<baseurl>/api/challenge/*` endpoint from within `ChallengeApiService.loadAll()`, retrieve the complete dataset, and return it to the calling component.

## Out of scope

* **Advanced Error Handling:** Detailed management of server timeouts or 500 errors.
* **Security:** Manually handling tokens or permissions (assuming global configuration).

## Acceptance Criteria

* The `ChallengeApiService.loadAll()` method successfully sends a request to the backend server.
* The service returns the real response value provided by the backend.
* If no response is received, the service returns the agreed-upon default value.
* The network request follows the project's standard API communication protocols.

soyjuandelgado:
Issue: https://github.com/IT-Academy-BCN/ita-challenges/issues/425

## Goal

Connect to `/api/challenge` and retrieve all challenges from the server.

## What has been done

- Added `HttpClient` to the project configuration in `app.config.ts`
- Called the GET endpoint in `loadAll()` using a hardcoded URL
- Since this is a **happy path**, in case of any error it returns `CHALLENGES_MOCK`
- Added tests for success and failure cases
- Communication tests


JordiMiravet (approved):
SGTM !

Andrescalvo22 (approved):
Overall, this is a solid implementation and the tests are well structured 👍
The use of HttpTestingController and the coverage of both success and error scenarios are especially good.

Note / suggestion:
I understand that some parts (like the hardcoded API URL and the use of CHALLENGES_MOCK) are in place because the backend integration is not yet available, which makes sense for this stage.

It might be helpful to briefly document this (or leave a TODO) so it’s clear that these should be replaced once the backend endpoints are fully integrated.

Overall, great work and good test coverage 👍

kenan-it-academy (commented):
Putting a pin on this one for a little bit cause I'm unsure about the tests

kenan-it-academy (approved):
Approved but backlog the test rework we talked about. We need a better way to make Http-dependant Observable tests that is:

1. More Readable
2. Less fragile

PR #437, opened 2026-04-28
refactor(auth): switch to GitHub OAuth login flow (#350)

JordiMiravet:
## Summary
Refactored the authentication flow from a simple mock-based implementation to a more structured mock flow simulating a GitHub OAuth login

## Changes
- Updated `AuthService` to replace Observable-based login mock with a service-driven mock user flow
- Added `fetchUser()` to simulate backend `/me` endpoint
- Simplified `AuthPageComponent` by delegating login logic to the service
- Added loading and error states in `ProfilePageComponent`
- Introduced `ngOnInit` to hydrate user state on profile load
- Improved profile template with loading/error/empty states handling

## Note
Still uses mocked data, but structured to mimic a real OAuth-based authentication flow for future backend integration

PR #448, opened 2026-04-29
feat(auth): add auth guard for protected routes (#349)

### LINKED ISSUE #349: [AUTH] - Add auth guard for protected routes
## Objective
Protect the `/profile` route so unauthenticated users are redirected to `/auth`, using `AuthService.getUser()` to check the session.

## Acceptance Criteria
- [x] `authGuard` exists in `core/guards/`
- [x] Redirects to `/auth` if no user is found
- [x] Allows navigation if user is present
- [x] Guard is applied in `profile.routes.ts`

## Prerequisites
- #365 merged

> These criteria may be refined as implementation evolves.

JordiMiravet:
## Summary
Added auth guard to protect the `/profile` route, redirecting unauthenticated users to `/auth`

## What's included
- `authGuard` created in `core/guards/`
- Guard applied to `/profile` route in `profile.routes.ts`
- Unit tests covering both authenticated and unauthenticated cases

## Notes
Only `/profile` is protected for now to avoid affecting other team members' routes. Other routes can be protected incrementally as the auth flow matures.

### DIFF
--- a/frontend/src/app/core/guards/auth-guard.spec.ts
+++ b/frontend/src/app/core/guards/auth-guard.spec.ts
@@ -1,5 +1,5 @@
 import { TestBed } from '@angular/core/testing';
-import { Router } from '@angular/router';
+import { Router, UrlTree } from '@angular/router';
 import { authGuard } from './auth-guard';
 import { AuthService } from '../../features/auth/data-access/auth-service';
 import { AuthUser } from '../../features/auth/models/auth-user.model';
@@ -19,7 +19,10 @@ describe('authGuard', () => {
         AuthService,
         {
           provide: Router,
-          useValue: { navigate: vi.fn() },
+          useValue: {
+            navigate: vi.fn(),
+            createUrlTree: vi.fn(() => ({} as UrlTree)),
+          },
         },
       ],
     });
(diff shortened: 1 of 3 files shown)


JordiMiravet:
This implementation was not strictly required for the current scope. I have added it for completeness, but it can be considered open for discussion whether it should be included in this iteration or deferred to a future one, depending on how the authentication flow evolves in upcoming tasks.

JordiMiravet:
Current implementation uses in-memory user state only:

```ts
return auth.getUser()
  ? true
  : router.createUrlTree(['/auth']);
```

This is intentionally synchronous because fetchUser() is not available in this branch yet,
otherwise it would break CI.

TODO (after #437 PR is merged):
Replace the current logic with an async guard that resolves the user from backend:

```ts
return auth.getUser()
  ? true
  : auth.fetchUser().pipe(
      map(user => {
        auth.setUser(user);
        return true;
      }),
      catchError(() => of(router.createUrlTree(['/auth'])))
    );
```

This will:

* Restore session on page reload
* Avoid unnecessary redirects for already authenticated users
* Centralize user fetching logic in the guard instead of components

Note:
Current version will redirect on refresh because user state is not persisted.

soyjuandelgado (approved):
LGTM! 😄

nereame96 (approved):
Looks good!

kenan-it-academy (approved):
This looks good but I'm assuming there's bonus time left in the sprint and everything is already finished? I recall this explicitly being mentioned as being out of scope for this sprint...

JordiMiravet:
> This looks good but I'm assuming there's bonus time left in the sprint and everything is already finished? I recall this explicitly being mentioned as being out of scope for this sprint...

You're right — we agreed in this morning’s daily to leave this PR for a future sprint, but with everything going on I forgot to update the status. Happy to move it to the next iteration if that’s preferable.

kenan-it-academy:
I mean, it's approved, you might as well merge lol

kenan-it-academy:
This is still approved...

PR #533, opened 2026-05-06
feat: user repository save

### LINKED ISSUE #512: [AUTH] Save a user during runtime
As a developer of the account microservice,
I want users to be saved while the application is running,
so that the system can persist new users without needing a database yet.
Acceptance criteria:

A user can be saved by username and role.
Saving a user that already exists overwrites the previous entry.
A test validates both scenarios: saving a new user and overwriting an existing one.

Out of scope: user lookup, database persistence, business logic.

eariasvalor:
## What
Adds `UserRepository` port and `InMemoryUserRepository` adapter with `save` functionality.

## Changes
- `User` — provisional record with `userName` (String) and `role` (Role) fields
- `Role` — provisional enum with `ADMIN`, `STUDENT`, `GUEST`
- `UserRepository` — output port interface with `save(User user)`
- `InMemoryUserRepository` — adapter implementing `save` via `ConcurrentHashMap`
- `InMemoryUserRepositoryTest` — covers save and overwrite scenarios

## Pending
- `User.java` and `Role.java` are temporary — belong to PR 0.1. Remove once merged.

## Depends on
- PR 0.1 — `User` and `Role` domain models


JordiCasas87 (changes requested):
I’ve noticed a couple of issues that should be fixed before merging:

Role.ADMIN is used in the tests, but ADMIN does not exist in the Role enum, so this currently breaks compilation/tests.
There also seems to be an inconsistency in the package structure between the repository implementation and the test package.

Apart from that, the implementation looks clear and aligned with the issue requirements.

eariasvalor:
Hi Jordi! Thanks for noticing it, my bad! It's all fixed now.

JordiCasas87 (approved):
Thanks for the updates. The requested changes have been applied and the implementation now looks aligned with the current in-memory repository approach. Looks good to me.

Rafadicandia (approved):
Esther: well done!

kenan-it-academy (approved):

PR #570, opened 2026-05-07
feat(auth-guard): add async session restoration fallback (#569)

### LINKED ISSUE #569: [AUTH] Improve auth guard with async session restoration
## Objective
Upgrade auth guard to support async session restoration from backend while keeping current in-memory fast path.

## Acceptance Criteria
- [x] Auth guard supports backend session restoration
- [x] User is not redirected to /auth on refresh when session exists
- [x] Unauthenticated users are redirected to /auth
- [x] Existing getUser fast-path still works
- [x] App compiles with no errors

## Prerequisites
- #349 merged

JordiMiravet:
## Summary
Updated auth guard to support backend session restoration when no in-memory user is available.

## What's changed
- Replaced pure in-memory check with async fallback using `fetchUser()`
- On missing in-memory user, attempt to restore session from backend
- Store retrieved user in memory using `setUser()`
- Redirect to `/auth` if session restoration fails

## Notes
This change allows authenticated sessions to persist across page refresh while keeping the original fast-path check for in-memory users. 
This is an update to the implementation introduced in PR #448

### DIFF
--- a/frontend/src/app/core/guards/auth-guard.ts
+++ b/frontend/src/app/core/guards/auth-guard.ts
@@ -1,5 +1,6 @@
 import { inject } from '@angular/core';
 import { CanActivateFn, Router } from '@angular/router';
+import { catchError, map, of } from 'rxjs';
 import { AuthService } from '../../features/auth/data-access/auth-service';
 
 export const authGuard: CanActivateFn = () => {
@@ -10,5 +11,11 @@ export const authGuard: CanActivateFn = () => {
 
   return user
     ? true
-    : router.createUrlTree(['/auth']);
+    : auth.fetchUser().pipe(
+        map(user => {
+          auth.setUser(user);
+          return true;
+        }),
+        catchError(() => of(router.createUrlTree(['/auth'])))
+      );
 };


andyrouge81 (approved):
Nice! 

The async fallback mechanism to restore the user session when there is no in-memory user makes the auth guard much more reliable after page refreshes while keeping the original fast-path behavior clean.

At this pace you are shipping PRs faster than Angular refreshes the page.

Apart from that, well scoped implemnetation and GJ!

eariasvalor (approved):
Makes sense ¯\_(ツ)_/¯

JordiMiravet:
> Nice!
> 
> The async fallback mechanism to restore the user session when there is no in-memory user makes the auth guard much more reliable after page refreshes while keeping the original fast-path behavior clean.
> 
> At this pace you are shipping PRs faster than Angular refreshes the page.
> 
> Apart from that, well scoped implemnetation and GJ!

> Makes sense ¯_(ツ)_/¯

Thanks both! Glad the fallback approach makes sense and improves reliability without touching the fast path. 
Thanks Andy, I like to be diligent with the implementation 🙂 Appreciate it.

nereame96 (approved):
Looks good for me!

kenan-it-academy (approved):

PR #598, opened 2026-05-07
feat: define answer progress states

### LINKED ISSUE #579: [CHALLENGE] Define the answer progress states
**Confirm the lifecycle of an answer:**

- not started
- in progress
- submitted
- approved
- failed review

YoussefEG09:
## Added AnswerStatus enum to define the lifecycle states of an answer:

- NOT_STARTED
- IN_PROGRESS
- SUBMITTED
- APPROVED
- FAILED_REVIEW

No unit tests added as there are no existing tests for the domain layer yet.

Closes #579

### DIFF
--- a/backend/challenge-service/src/main/java/com/itachallenges/challengeservice/catalog/domain/valueobject/AnswerStatus.java
+++ b/backend/challenge-service/src/main/java/com/itachallenges/challengeservice/catalog/domain/valueobject/AnswerStatus.java
@@ -1,0 +1,9 @@
+package com.itachallenges.challengeservice.catalog.domain.valueobject;
+
+public enum AnswerStatus {
+    NOT_STARTED,
+    IN_PROGRESS,
+    SUBMITTED,
+    APPROVED,
+    FAILED_REVIEW
+}


Andrescalvo22 (approved):
Looks good and clear, approved!

eariasvalor (approved):
Nice one, GJ!

yvangabrieli (approved):
Clean job. Approve GJ Youssef

andyrouge81 (changes requested):
One small architectural doubt from my side:
while reviewing the current structure, I noticed we already have a `SubmissionStatus` enum inside the `submission` domain, but this PR introduces a new `AnswerStatus` under `catalog`.

Maybe this happened because of the naming used in the issues/board (`answer`, `challenge solution`, `submission`), which can be a bit confusing at this stage of the project 

The reason I mention it is because currently the technical domain seems to revolve around `Submission` (`Submission`, `SubmissionId`, `SubmissionStatus`, `SubmissionFinished`, etc.), and having both `SubmissionStatus` and `AnswerStatus` could introduce duplicated concepts in future tasks or in-progress issues.

Perhaps it would be good to decide which naming we want to keep (`Submission` vs `Answer`) and explain briefly the reason behind that decision, so the rest of the current and future issues can stay aligned with the same ubiquitous language across the project.

What do you think?

PR #606, opened 2026-05-07
doc(challenge): define shared structure for student challenge submission

### LINKED ISSUE #578: [CHALLENGE] Define the information stored in a student answer
**Agree on the minimum information needed when a student responds to a challenge:**

- related challenge
- student
- answer content
- current progress state

yvangabrieli:
## Overview

This PR adds documentation for the student submission structure used in the challenge flow.

The goal is to make sure the whole team shares the same understanding of what information a student answer contains before continuing development.

## Why this is important

This documentation helps avoid inconsistencies between frontend and backend implementation by defining:
- what information must be stored
- the basic rules of a student answer
- what is included and excluded from the MVP scope

Having this agreed early reduces confusion during development and future integrations.

## What is included

- Definition of a student submission
- Required information/fields
- Basic validation rules
- Acceptance criteria
- Out of scope clarification

## Notes

The backend structure already contains these fields.  
This PR mainly documents and confirms the agreed model so the team can work consistently from the same reference.

### DIFF
--- a/backend/challenge-service/docs/submission-entity.md
+++ b/backend/challenge-service/docs/submission-entity.md
@@ -1,45 +1,69 @@
-# Title: [CHALLENGE] Agree on the information saved in a Submission (student answer)
+# Submission Entity — Student Answer
+
+> **This document is subject to change.** This is an early version of the project and some concepts may evolve as development progresses. Any changes should be discussed and agreed on by the team before implementation.
+
+---
 
 ## Overview
-Before starting development, we need to agree on what information should be saved when a student works on a challenge.  
+
+Before starting development, we need to agree on what information should be saved when a student works on a challenge.
+
 This helps everyone follow the same structure and avoids confusion later.
 
+---
+
 ## What is a submission?
-A submission (student answer) is created when a student starts a challenge.  
+
+A submission (student answer) is created when a student sends their answer to the mentor for review.
+
 It is connected to:
 - one student
 - one challenge
 
-It also keeps track of the student’s work and progress.
+It also keeps track of the student's work and progress.
+
+---
 
 ## Information that should be saved
 
-| Information | Meaning |
-|---|---|
-| ID | Unique reference for the answer |
-| Challenge ID | Which challenge the answer belongs to |
-| Student ID | Which student created the answer |
-| Code / Solution | The student’s written solution |
-| Status | Current progress state |
-| Created date | When the answer was first created |
-| Last updated date | When the answer was last changed |
+| Information      | Meaning                                  |
+|------------------|------------------------------------------|
+| ID               | Unique reference for the answer          |
+| Challenge ID     | Which challenge the answer belongs to    |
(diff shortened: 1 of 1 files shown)


JordiMiravet (approved):
LGTM! The documented submission entity matches the current backend structure and clearly defines the expected fields and constraints

andyrouge81 (changes requested):
Good job! The documentation is clear and useful to align frontend and backend before continuing with the implementation.

Just two small points to clarify:

* About *"when a student starts a challenge"*: from my understanding, the submission should not be created when the student starts the challenge, but when the student completes the challenge and wants to send it to the mentor for correction. Maybe this detail will be refined later, but I think it is good to clarify it early.

* About *"The solution field must be filled before saving or submitting"*: I think *before submitting* fits better, because a student could save progress without filling the solution field yet.

Also, I think it could be useful to mention that this documentation is always subject to future changes and refinements, since these are still early versions of the project and some concepts may evolve during development.

Apart from these small details, everything else looks good to me. If we need to clarify these concepts, better now than later in development. GJ!

JordiCasas87 (changes requested):
Looks good overall.

I also agree with the clarification about creating the submission when the student submits the challenge, and allowing progress to be saved before the solution field is completed.

Apart from those small details, everything else looks clear to me.

EstherBlacksmith (changes requested):
Hi! I'm not sure if the student can only send one answer to a challenge, or maybe it's possible to allow the student to send various attempts to resolve the challenge. I think this must be discussed with the product owner.🤷‍♀️
Having said that, very good documentation work!!

yvangabrieli:
Thanks for the feedback! You're right on both points.
I've updated the doc to clarify that the submission is created when the student submits their answer, not when they open the challenge. I've also changed the rule so that the solution field is only required before submitting, not before saving a draft — that makes more sense for the "save progress" flow.
Good catch on adding a note about the doc being subject to change — added that too.

yvangabrieli:
Great question — you're right that this needs a product owner decision. For now I've kept the rule as "one answer per challenge" since that's the simplest assumption to start with, but I've added a note flagging it as an open question to confirm with the PO. Thanks for raising it early!

Andrescalvo22 (approved):
Looks Good!

nereame96 (approved):
Looks good!

kenan-it-academy (approved):

andyrouge81 (approved):
Thanks for updating the documentation, Yvan.

The points raised during the review are much clearer now, especially around submission creation, draft saving flow, and the open questions that still need PO validation.

I also think this kind of documentation PR is really important for the team, because it helps future iterations stay aligned and gives everyone a shared reference before implementation starts growing. GJ!

EstherBlacksmith (approved):

JordiCasas87 (approved):
Looks good for me!! Thanks!

Issue #580, opened 2026-05-07
[CHALLENGE] Allow students to save a draft answer


yvangabrieli:
**Students can save their work without submitting it yet.**

*Expected behavior:*

- answer is stored
- progress state becomes “In Progress”

Issue #577, opened 2026-05-07
[CHALLENGE] As a student, I want to submit challenge solutions and confirm the submission


yvangabrieli:
## Overview

This feature introduces the first version of the student answer flow for challenges. Students should be able to start an answer, save their progress as a draft, and submit their final response when ready.

The system should track the progress of the answer through simple states such as:

- [ ] not started
- [ ] in progress
- [ ] submitted

## Out of Scope

This phase does not include:

- editing or deleting answers
- full mentor review workflows
- grading
- notifications
- advanced permissions or rules
- complete answer management features

This phase focuses only on the core experience needed to support challenge submissions and prepare the foundation for future mentor review features.

PR #621, opened 2026-05-10
feat(challenge): validate student submissions in happy path

andyrouge81:
## Description

This PR introduces the first MVP implementation for the student submission happy path.

## What is included

* Initial endpoint:

  * `POST /api/v1/submissions`

* Basic request mapping for:

  * `challengeId`
  * `userId`
  * `content`

* Temporary support for empty answer content during the MVP phase

* Minimal controller tests for:

  * request mapping
  * empty content handling

## Current MVP scope

For this first iteration:

* the flow currently focuses only on the happy path
* answer content can temporarily be empty
* only one save/submit action exists
* strict validation and workflow rules are left for future iterations

## Notes

This implementation is intended as a small incremental first step for the submission flow.

Some parts of this subissue may still evolve depending on related tasks such as answer structure, workflow states, and submission behavior.

Closes #582
Related to #577

PR #622, opened 2026-05-11
Feat/finalize submission 581

### LINKED ISSUE #581: [CHALLENGE] Allow students to submit an answer for review
**Students can officially send their final answer for mentor review.**

*Expected behavior:*

- answer is stored/submitted
- progress state becomes “Submitted”

yvangabrieli:
# What this PR does

Implements the finalize submission flow so students can officially send their answer for mentor review.

---

# Changes included

- Added `POST /api/v1/submissions/finalize` endpoint
- Implemented `Submission.finalize()` with state guard validation
  - cannot finalize if submission is already `SUBMITTED`
- Renamed `SubmissionStatus.FINAL` → `SUBMITTED` to better match the user story language
- Updated `FinalizeSubmissionRequest` and `FinalizeSubmissionCommand` to use `List<String> content`
  - aligned with the structure introduced in #580
- Added controller tests

---

# Note for reviewers

The rename from `FINAL` → `SUBMITTED` will create a small merge conflict once PR #580 is merged.

At that point, references to:

```java
existsFinalSubmission()
```

will need to be updated accordingly.

---

Closes #581  
Related to #577


eariasvalor:
Hi Yvan!
Nice work, the only thing is maybe this PR is too large, it could be split into smaller ones. Please consider discarding the implementation of the handler and all the service layer, which could be left for future sprints. This should make the PR shorter.

yvangabrieli:
Thanks for the heads up Esther so now the finalize submission endpoint so students can officially send their answer for mentor review. The HTTP layer is fully set up. The business logic (use case handler and domain transition) is left for the next sprint.

backend/challenge-service/src/main/java/com/itachallenges/challengeservice/submission/infrastructure/adapter/in/web/controller/SubmissionController.java
```diff
@@ -22,7 +30,19 @@ public SubmissionController(FinalizeSubmissionUseCase finalizeSubmissionUseCase,
...
+    @PostMapping("/finalize")
+    public ResponseEntity<Void> finalize(@RequestBody FinalizeSubmissionRequest request) {
+        try {
+            List<String> content = request.content() == null ? List.of() : request.content();
+            FinalizeSubmissionCommand command = new FinalizeSubmissionCommand(
+                    request.challengeId(),
+                    request.userId(),
+                    content
+            );
+            finalizeSubmissionUseCase.execute(command);
```
eariasvalor: I see you still call the execute method from the UseCase, if we are removing the application layer, the endpoint should have the business logic for now

backend/challenge-service/src/test/java/com/itachallenges/challengeservice/submission/infrastructure/adapter/in/web/controller/SubmissionControllerTest.java
```diff
@@ -1,29 +1,102 @@
...
 
-    @Test
-    void shouldFinalizeSubmission() {
-        // TODO: implement
-    }
+    @Autowired
+    private ObjectMapper objectMapper;
+
+    @MockBean
+    private FinalizeSubmissionUseCase finalizeSubmissionUseCase;
```
eariasvalor: Same here, the usecase shouldn't be injected or used in the test either

eariasvalor (commented):
Hi Yvan!
There are stilll injections of the usecase in the controller. The endpoint should be able to manage all logic without the UseCase class. Also, the tests should inject the repository for now, not the service.

yvangabrieli:
Done! Removed the use case injection from both the controller and the test. The controller now injects the repository directly and handles the duplicate check inline. Business logic TODO comment left for the next sprint.

backend/challenge-service/src/main/java/com/itachallenges/challengeservice/submission/infrastructure/adapter/in/web/controller/SubmissionController.java
yvangabrieli: Done! Thank you Esther

backend/challenge-service/src/test/java/com/itachallenges/challengeservice/submission/infrastructure/adapter/in/web/controller/SubmissionControllerTest.java
yvangabrieli: Done! Thank you Esther

andyrouge81 (changes requested):
I reviewed the changes again and I think the main issue here comes from a small mix-up between our latest changes. Yesterday I reverted my part back to `String code` after the tech lead suggestion, but I accidentally dragged part of your `content` implementation while rebasing/continuing the work on top of it.

For this MVP phase, I think keeping `String code` is more aligned with how `Submission` is currently represented in the domain, and it also keeps the flow simpler for now.

So probably we should just align the remaining pieces (`FinalizeSubmissionCommand`, request DTO, controller and tests) back to `code` to keep everything consistent again.

Apart from that, the endpoint flow and tests look clear. GJ!

yvangabrieli:
Thank you for the review and for catching that. You were right, it came from the mix-up during the rebase after reverting the implementation back to `String code`.

I’ve now refactored everything to keep it consistent again with the current domain representation, including the remaining pieces (`FinalizeSubmissionCommand`, request DTO, controller, and tests).

Appreciate the detailed feedback 🙌

andyrouge81 (approved):
Good job with the latest changes! The PR feels much cleaner now and the `/api/challenge/...` prefix fits better with the current convention.

I think what you implemented is totally valid. Maybe just as a small MVP consideration, since we are trying to keep things as lightweight as possible in this phase, we could even simplify a bit more and remove the remaining `FinalizeSubmissionCommand`.

Apart from that, the happy path and the tests look good to me. GJ!

jrffruit518-enrich (approved):
Looks good to me now.

Thanks for aligning the implementation back to String code and cleaning up the remaining inconsistencies after the rebase. The endpoint flow and tests are clear, and the current MVP scope feels consistent with the rest of the project.

kenan-it-academy (changes requested):
Isn't this PR entirely out of scope???  Am I crazy?

yvangabrieli:
Hey Kenan! You were totally right — the previous version was indeed out of scope. The controller was only checking for duplicates but never actually saving anything or transitioning any state, so the core requirement wasn't being met.

I've reworked the PR to properly implement the feature as described in the issue:

- `Submission.finalize()` now guards against re-submission and transitions the status to `SUBMITTED`
- `FinalizeSubmissionService` wires the use case: finds the submission, calls finalize, and saves it back
- `InMemorySubmissionRepository` now implements `findByUserAndChallenge` and `existsFinalSubmission`
- `SubmissionController` now delegates to the use case instead of calling the repository directly
- Controller tests updated to mock the use case layer and cover the 404 case too

Every file changed is required by the same feature — it's a thin vertical slice so I kept it as one PR. Happy to discuss if you think it should be structured differently!

andyrouge81 (changes requested):
I think the PR is much more aligned now with the actual finalize submission flow. 

I only noticed two small things before approving:

- In `SubmissionControllerTest`, `ObjectMapper` seems to be duplicated at the end of the class, which may break compilation.
- The test `shouldFinalizeWithEmptyCodeWhenCodeIsMissing` expects empty code handling, but the finalize endpoint currently passes `request.code()` directly instead of normalizing `null `to `""` like the draft endpoint does.

Apart from that, the scope and implementation feel much clearer now. GJ!

yvangabrieli:
Good catch — this was introduced while resolving conflicts after merging changes from branch #621.

During the conflict resolution process, part of the draft submission test block remained in the file accidentally, which left duplicated fields (`ObjectMapper`) and misplaced test code inside the class.

I cleaned up the leftover merge artifacts and also aligned the finalize endpoint behavior with the draft endpoint by normalizing `null` code values to an empty string.

andyrouge81 (changes requested):
Nice progress on this PR, and thanks for addressing the previous comments so quickly.

I just noticed one small thing: in the test the `null` code becomes `""`, but in the controller I still see `request.code()` directly.

Not sure if that was intentional for the finalize flow or just something left during the refactor.

andyrouge81 (approved):
GJ! Yvan.

I reviewed the latest changes and the PR feels much cleaner and more consistent now. The finalize flow is properly connected, the controller/tests are aligned again, and the merge conflict leftovers seem resolved too.

Thanks for addressing all the feedback so quickly!

kenan-it-academy:
@yvangabrieli @andyrouge81 we can talk about this later, but as I understand, submitting is the only thing in scope, with no distinction between challenges being "finalized" or not.

yvangabrieli:
You're right — I removed the finalize step completely.
The PR now only handles the submit flow using the existing POST /api/challenge/submissions endpoint, where answers are saved as submitted.

I also added a simple GET /{id} endpoint so we can easily check that the data is being saved correctly.

kenan-it-academy:
Why is Post and Get /id lumped together in the same PR?

yvangabrieli:
You're right, the GET endpoint is out of scope for this issue. I added it thinking it would help demo the persistence but it doesn't belong here. I'll remove it and keep only the handler implementation that stores the submission with SUBMITTED status. The test covers the behaviour already.

backend/challenge-service/src/test/java/com/itachallenges/challengeservice/submission/infrastructure/adapter/in/web/controller/SubmissionControllerTest.java
```diff
@@ -44,11 +47,32 @@ void shouldSubmitWithEmptyCodeWhenCodeIsMissing() throws Exception {
...
+
+        ArgumentCaptor<SaveDraftSubmissionCommand> captor =
+                ArgumentCaptor.forClass(SaveDraftSubmissionCommand.class);
+        verify(saveDraftSubmissionUseCase).execute(captor.capture());
         assertThat(captor.getValue().code()).isEmpty();
     }
-}
+
+    @Test
+    void shouldReturn404WhenSubmissionNotFound() throws Exception {
```
kenan-it-academy: This test shouldn't be here, GET was removed

kenan-it-academy (changes requested):

backend/challenge-service/src/test/java/com/itachallenges/challengeservice/submission/infrastructure/adapter/in/web/controller/SubmissionControllerTest.java
yvangabrieli: Good catch. you are completely right. Test has been removed.

kenan-it-academy: @yvangabrieli still not removed, forget to push?

yvangabrieli: Now it is done. Pardon me for that.

kenan-it-academy (approved):

PR #643, opened 2026-05-12
feat(auth): implement create endpoint for users with upsert logic

### LINKED ISSUE #532: [AUTH] Allow creating new user or update existing user role
## Description
No input validations.

## Acceptance Criteria
- [ ] Tests passing.

AlbertMedina:
## Description
Implemented the `create` endpoint to handle user creation and role updates (upsert logic). The endpoint has been moved from the `UserController` to the `AuthController` to better align with the application's authentication flow.

## Changes
- `UserController` &  `UserControllerTest`: removed the `POST /users` endpoint and its corresponding unit tests.
- `AuthController` &  `AuthControllerTest`: added `POST /auth/register` endpoint with upsert logic implementation with its corresponding unit tests.
- `User`: added `withRole` method to support immutable role updates.
- `UserRepository` &  `InMemoryUserRepository`: changed the `save` method return type from `void` to `User` to support returning the saved user.

## Implementation details
We'll need to double-check that everything works as expected once the other Auth PRs are merged. Additionally, a future refactor will be necessary to rename `username` to `githubId` across the application, as we are now saving the GitHub ID instead of the username.


JordiCasas87 (approved):
Looks good to me. The upsert flow is clear, the endpoint move to AuthController makes sense, and the tests cover the main scenarios correctly. Nice work!

eariasvalor (approved):
Everything looks OK to me, nice work!

backend/account-service/src/test/java/com/ita/challenges/account/infrastructure/adapter/web/in/AuthControllerTest.java
```diff
@@ -17,6 +32,75 @@ class AuthControllerTest {
...
+    private UserRepository repository;
+
+    @MockBean
+    private RestTemplate restTemplate;
+
+    @Autowired
+    private ObjectMapper objectMapper;
+
+    @Test
+    void create_shouldReturn201WhenUserIsNew() throws Exception {
```
kenan-it-academy, line 45: All three of these tests (and their implementations) should be separate PRs

kenan-it-academy (changes requested):

AlbertMedina:
Closing this PR to break it down into smaller, more manageable PRs as suggested. This will make the review process easier. Thanks for the feedback!

PR #614, opened 2026-05-08
(Ticketing): define update method in TicketRepository

jrffruit518-enrich:
### Summary
add updateTicket(Ticket ticket) to TicketRepository

add a provisional override in InMemoryTicketRepository

add a simple test for the provisional implementation

### Note
This PR only defines the contract required to update a ticket and leaves the real storage logic for the following sub-issue.

closes #481

PR #609, opened 2026-05-07
feat(ticketing): retrieve existing tickets from storage

JordiCasas87:
Closes #477

## Summary
- implement `findAllByUserId(String userId)` in `InMemoryTicketRepository`
- fulfill the contract defined in `TicketRepository`
- add a simple test covering the empty-list case

## Note
This PR only covers the storage-side retrieval logic for user tickets and does not include endpoint logic.

Only the empty-list case is covered in this PR. Testing actual retrieval by user depends on having ticket persistence populated through the corresponding ticket creation flow, which is being handled in a separate issue.

PR #645, opened 2026-05-12
feat: add ticket repository contract

### LINKED ISSUE #497: [TICKETING] [ENDPOINT] Prepare the ticket create entry point
## Description
Create the REST entry point for the ticket create operation within the `TicketController`.

Following the project's simplified architecture, this task involves setting up the controller method that will eventually call the repository directly. The goal is to establish the routing and request handling (request body) without implementing the full creation logic yet.

### Acceptance Criteria

- [ ] The `TicketController` includes a POST mapping for creating tickets.
- [ ] The endpoint is configured to receive the ticket creation data in the request body.
- [ ] The controller method skeleton is prepared to call the `TicketRepository`.
- [ ] A basic integration test is included to verify the endpoint is reachable (e.g., returning a 201 Created or 501 Not Implemented status).
- [ ] No persistence logic or data mapping is implemented in this sub-issue.
- [ ] A simple test is included.

Spook242:
Closes #497

### Description
This PR adds the `TicketRepository` interface to define the contract for ticket persistence in the domain layer.

### ⚠️ Note to reviewers
As requested, I have split my original branch into three separate PRs to follow the "1 issue = 1 PR" workflow.

### DIFF
--- a/backend/account-service/src/main/java/com/ita/challenges/account/domain/port/out/TicketRepository.java
+++ b/backend/account-service/src/main/java/com/ita/challenges/account/domain/port/out/TicketRepository.java
@@ -2,10 +2,6 @@ package com.ita.challenges.account.domain.port.out;
 
 import com.ita.challenges.account.domain.model.Ticket;
 
-import java.util.List;
-
 public interface TicketRepository {
-    List<Ticket> findAllByUserId(String userId);
-
-    Ticket updateTicket(Ticket ticket);
+    Ticket save(Ticket newTicket);
 }
--- a/backend/account-service/src/main/java/com/ita/challenges/account/infrastructure/adapter/web/out/persistence/InMemoryTicketRepository.java
+++ b/backend/account-service/src/main/java/com/ita/challenges/account/infrastructure/adapter/web/out/persistence/InMemoryTicketRepository.java
@@ -4,25 +4,16 @@ import com.ita.challenges.account.domain.model.Ticket;
 import com.ita.challenges.account.domain.port.out.TicketRepository;
 import org.springframework.stereotype.Repository;
 
-import java.util.ArrayList;
-import java.util.List;
+import java.util.HashMap;
 import java.util.Map;
-import java.util.concurrent.ConcurrentHashMap;
 
 @Repository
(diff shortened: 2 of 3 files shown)


JordiCasas87 (changes requested):
Thanks for the changes. I think this PR is still blocked for now because Sonar is failing and the project does not compile yet.

Right now only the repository contract was updated, but the temporary implementation was not adapted to the new method signature. Since the implementation is missing, compilation breaks.

Could you also update the corresponding repository implementation so the project builds correctly?

You could follow the same approach used in PRs #609 and #614, where the temporary in-memory implementation was also adapted to keep the project compiling while the feature is still in progress.

Also, once this is solved, it would be great to add at least a small/minimal test covering the new behavior.

Spook242:
I split it too much and broke the interface implementation rule. I've just merged the adapter implementation into this PR to fix the build and Sonar issues.

JordiCasas87 (changes requested):
The PR adds the `save` method correctly, but it also removes existing methods from `TicketRepository` and `InMemoryTicketRepository` (`findAllByUserId` and `updateTicket`).

Those methods are part of work already implemented by other teammates during this sprint and should not be removed in this scope.

Additionally, the repository changes from `ConcurrentHashMap` to `HashMap`, which removes the thread-safe behavior that already existed.

I would suggest restoring the branch from `develop` (or recovering the removed code) and then implementing only the required `save` method for this issue.

It would also be good to add at least a minimal test covering the new `save` behavior.

Spook242:
Minimal test is green and ready for review!

JordiCasas87:
Thanks for fixing the build issues and adding the minimal test.

However, the main blocking point is still pending. The PR continues removing `findAllByUserId` and `updateTicket` from the repository contract and implementation (these methods are already part of work implemented by other developers during this sprint).

I think the best approach here would be to restore/rebase the branch from the current `develop` state and then apply only the changes required for this issue (`save`), so no existing functionality is accidentally removed.

Spook242:
Hey guys, my Git history got really messy while resolving conflicts. To make it easier to review, I'm closing this PR and I'll open a new, much cleaner one.

PR #649, opened 2026-05-12
feat(auth): setup for create/update user endpoint

### LINKED ISSUE #532: [AUTH] Allow creating new user or update existing user role

AlbertMedina:
## Description
This PR prepares the initial setup required to implement the user creation and update endpoint.

## Changes
- `UserRepository` &  `InMemoryUserRepository`: changed the `save` method return type from `void` to `User` to support returning the saved user.
- `InMemoryUserRepositoryTest`: added test for `save` method.
- `User`: added `withRole` method to support immutable role updates.
- `UserTest`: added test for `withRole` method.
- `AccountApplication`: added the `RestClient` bean to enable communication with GitHub API.
- `GitHubUserResponse`: created a new DTO to capture the GitHub ID from the GitHub API response.ç

## Implementation details
This PR provides the foundational setup; the endpoint logic will be delivered in subsequent PRs.

### DIFF
--- a/backend/account-service/src/main/java/com/ita/challenges/account/AccountApplication.java
+++ b/backend/account-service/src/main/java/com/ita/challenges/account/AccountApplication.java
@@ -3,7 +3,7 @@ package com.ita.challenges.account;
 import org.springframework.boot.SpringApplication;
 import org.springframework.boot.autoconfigure.SpringBootApplication;
 import org.springframework.context.annotation.Bean;
-import org.springframework.web.client.RestTemplate;
+import org.springframework.web.client.RestClient;
 
 @SpringBootApplication
 public class AccountApplication {
@@ -13,7 +13,7 @@ public class AccountApplication {
 	}
 
     @Bean
-    public RestTemplate restTemplate() {
-        return new RestTemplate();
+    public RestClient restClient() {
+        return RestClient.create();
     }
 }
--- a/backend/account-service/src/main/java/com/ita/challenges/account/AccountApplication.java
+++ b/backend/account-service/src/main/java/com/ita/challenges/account/AccountApplication.java
@@ -2,6 +2,8 @@ package com.ita.challenges.account;
 
 import org.springframework.boot.SpringApplication;
 import org.springframework.boot.autoconfigure.SpringBootApplication;
+import org.springframework.context.annotation.Bean;
+import org.springframework.web.client.RestTemplate;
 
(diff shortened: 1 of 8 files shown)


backend/account-service/src/main/java/com/ita/challenges/account/AccountApplication.java
```diff
@@ -10,4 +12,8 @@ public static void main(String[] args) {
 		SpringApplication.run(AccountApplication.class, args);
 	}
 
+    @Bean
+    public RestTemplate restTemplate() {
```
eariasvalor: Please consider using WebClient directly instead of restTemplate, which I think is deprecated since Spring Boot3

eariasvalor (changes requested):
Great job! Please check out my comment about restTemplate and also, maybe we could have a couple of tests for InMemoryUserRepository.save() and User.withRole()?

backend/account-service/src/main/java/com/ita/challenges/account/AccountApplication.java
AlbertMedina: Hi Esther! Thanks for the feedback! I've switched to RestClient instead. This way we modernize the code without switching to reactive programming, which may be out of scope for now. We can definitely consider a fully reactive approach in the future!

AlbertMedina:
Thank you Esther! Tests have been added and RestTemplate has been replaced by RestClient.

backend/account-service/src/main/java/com/ita/challenges/account/AccountApplication.java
eariasvalor: I think this is a good compromise. Thanks for finding the sweet spot between legacy code and overengineering!

eariasvalor (approved):
Good job!

andyrouge81 (approved):
The repository changes, the new `withRole()` method, the GitHub response DTO, and the `RestClient` setup all fit well as groundwork for the upcoming create/update user flow. The added tests also help keep the new behavior covered without overcomplicating the PR.

It feels like a solid first phase that leaves room for future iterations to continue building the authentication and role flow naturally.

From my side, GJ!

soyjuandelgado (approved):
Looks good!

kenan-it-academy (approved):

PR #650, opened 2026-05-12
feat(auth): implement create endpoint

### LINKED ISSUE #532: [AUTH] Allow creating new user or update existing user role

AlbertMedina:
## Description
Implemented the `create` endpoint to handle user creation. The endpoint has been moved from the `UserController` to the `AuthController` to better align with the application's authentication flow.

## Changes
- `UserController` &  `UserControllerTest`: removed the `POST /users` endpoint and its corresponding unit tests.
- `AuthController` &  `AuthControllerTest`: added `POST /auth/register` endpoint with user creation implementation with its corresponding unit test.

## Implementation details
Upsert logic will be implemented in subsequent PRs.

## Depends on
PR https://github.com/IT-Academy-BCN/ita-challenges/pull/649


eariasvalor (approved):
Good job!

andyrouge81 (changes requested):
Moving the endpoint into `AuthController` makes much more sense conceptually, and separating the initial user creation from the future update flow also keeps the scope easier to follow.

I also understand that at this stage there is still no dedicated Service layer, so having more logic temporarily inside the controller feels normal for an early iteration like this. It is clear that the structure is still evolving.

A couple of small observations:

`UserControllerTest` now looks empty after removing the old endpoint test, so maybe it could be removed entirely.
I was also wondering if `GitHubUserResponse` might fit better in an `out` package since it represents a response coming from an external API client rather than an incoming request to our own API.

The PR is a bit large, but I can also see that part of the size comes from the previous setup work and the dependency with #649, so the overall direction still feels coherent.

Looks good overall. GJ!

AlbertMedina:
Hi Andrés! Thanks for the feedback! You are totally right about `GitHubUserResponse`; I’ve moved it to the `out` package as suggested. Regarding the empty `UserControllerTest`, I decided to keep it because we’ll need it soon for future endpoints (users CRUD) and I thought it was better to leave the structure ready.

andyrouge81 (approved):
Thanks for the changes, especially moving GitHubUserResponse to the out package. The PR looks much cleaner now.

I only noticed that UserController is now left empty. Maybe this is intentional for future work or upcoming endpoints, so not a blocker from my side, just something to keep in mind.

Other than that, GJ!

AlbertMedina:
Please note that in `AuthControllerTest`, the diff shows the entire file as modified even though I only added the new endpoint test. This happened after merging with the previous auth PRs. I am not sure why this is occurring, but the existing code remains unchanged.

kenan-it-academy (changes requested):
@AlbertMedina PR should point to develop

kenan-it-academy (approved):

PR #662, opened 2026-05-13
feat(challenge): implement getById in challenge service #661

### LINKED ISSUE #661: [CHALLENGE] - Get challenge by Id
Description: Implement in challengeService the function GetById which return a a challenge filtered by Id.

nereame96:
[Task here](https://github.com/IT-Academy-BCN/ita-challenges/issues/661)

Description: Implement in challengeService the function GetById which return a a challenge filtered by Id.

### DIFF
--- a/frontend/src/app/features/challenges/services/challenge.service.spec.ts
+++ b/frontend/src/app/features/challenges/services/challenge.service.spec.ts
@@ -1,5 +1,5 @@
 import { TestBed } from '@angular/core/testing';
-import { of } from 'rxjs';
+import { firstValueFrom, of } from 'rxjs';
 import { vi } from 'vitest';
 import { ChallengeService } from './challenge.service';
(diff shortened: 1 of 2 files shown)


soyjuandelgado (approved):
Looks good!

Nice job!!

andyrouge81 (approved):
Nice work! I think this is a good approach for the current MVP phase and it helps frontend continue moving without waiting for a dedicated backend endpoint.

I like that the service is already prepared around a future `getById(id)` flow, because later it should be very easy to replace the current mock implementation with a real backend GET request without changing the components that consume it.

The current `loadAll() + find()` solution feels simple, clean and practical for this stage of the project, while still leaving a good base for future backend integration.

GJ!

JordiMiravet (approved):
LGTM!

kenan-it-academy (approved):

PR #603, opened 2026-05-07
[AUTH] SecurityConfig changes - OAuth2 and endpoint permissions   https://github.com/IT-Academy-BCN/ita-challenges/issues/452

EstherBlacksmith:
This pr is a split from the origial (to large) pr [https://github.com/IT-Academy-BCN/ita-challenges/issues/452](https://github.com/IT-Academy-BCN/ita-challenges/pull/url)
Include this commits fd42f2f, 6e9bc02.
Those file have the next comment/reviews in the original pr: 
- https://github.com/IT-Academy-BCN/ita-challenges/pull/455#discussion_r3197946978  
- https://github.com/IT-Academy-BCN/ita-challenges/pull/455#pullrequestreview-4240085061

PR #674, opened 2026-05-13
fix: correct GitHub OAuth redirect URI and post-login redirect to frontend

eariasvalor:
Fix GitHub OAuth authentication redirect flow

Three misconfigurations were causing the GitHub OAuth flow to fail:

1. loginPage: changed from /api/account/oauth2/authorization/github to /auth. This is a browser redirect to the frontend login page, not a backend route, so it must not carry the /api/account/ prefix.
2. defaultSuccessUrl: changed from /api/account/profile to /profile. Same reasoning: after a successful login, the browser should land on the frontend profile page, which nginx routes to the frontend service. Using the /api/account/ prefix was sending the user to a non-existent backend endpoint.
3. redirect-uri in application.yaml: added explicit configuration so Spring sends http://localhost:8080/login/oauth2/code/github to GitHub as the callback URL, matching what GitHub actually redirects to after authorization.

### DIFF
--- a/backend/account-service/src/main/java/com/ita/challenges/account/infrastructure/config/SecurityConfig.java
+++ b/backend/account-service/src/main/java/com/ita/challenges/account/infrastructure/config/SecurityConfig.java
@@ -27,7 +27,7 @@ public class SecurityConfig {
                         .redirectionEndpoint(redirection -> redirection
                                 .baseUri("/api/account/login/oauth2/code/**")
                         )
-                        .defaultSuccessUrl("/api/account/profile", true)
+                        .defaultSuccessUrl("/profile", true)
                 );
         return http.build();
     }
--- a/backend/account-service/src/main/java/com/ita/challenges/account/infrastructure/config/SecurityConfig.java
+++ b/backend/account-service/src/main/java/com/ita/challenges/account/infrastructure/config/SecurityConfig.java
@@ -20,7 +20,7 @@ public class SecurityConfig {
                         .anyRequest().authenticated()
                 )
                 .oauth2Login(oauth2 -> oauth2
-                        .loginPage("/api/account/oauth2/authorization/github")
+                        .loginPage("/auth")
                         .authorizationEndpoint(authorization -> authorization
                                 .baseUri("/api/account/oauth2/authorization")
                         )
--- a/backend/account-service/src/main/resources/application.yaml
+++ b/backend/account-service/src/main/resources/application.yaml
@@ -6,6 +6,7 @@ spring:
       client:
         registration:
           github:
+            redirect-uri: "{baseUrl}/api/account/login/oauth2/code/{registrationId}"
             client-id: ${GITHUB_CLIENT_ID}
             client-secret: ${GITHUB_CLIENT_SECRET}
             scope: user:email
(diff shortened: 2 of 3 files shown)


JordiCasas87 (changes requested):
I think there may be a small inconsistency in the documentation.

In `application.yaml` and `SecurityConfig`, the configured redirect URI points to:

`/api/account/login/oauth2/code/github`

But in `GitHub-Login&Callback-Setup.md` the return URL is documented as:

`/api/account/auth/github/github`

Since the auth flow is working, I think the implementation is correct and only the documentation path may need to be updated to match the configured redirect endpoint.

eariasvalor:
Hello, Jordi! You are absolutely right, I fixed the documentation now!

JordiCasas87 (approved):
Looks good to me now, thanks for the quick fix and for updating the documentation to match the configured redirect endpoint.

soyjuandelgado (approved):
Nice job!

Thank you for the fixing!!

EstherBlacksmith (changes requested):
I'm not sure about this, but I think Kenan asks for not living any point outside the API/account in this comment https://github.com/IT-Academy-BCN/ita-challenges/pull/603/changes/BASE..d5c7de388a07f3233813d02cac1f58a60698b8ed#r3226091751

eariasvalor:
Hi @EstherBlacksmith, the urls that don't have api/account are not backend endpoints, but frontend routes, as the description of the PR points out. This is the reason the flow didn't work, because github needs to handle the frontend routes, not point to non-existing backend endpoints.

EstherBlacksmith (approved):
Thanks for the explanation. Good job on fixing it!

backend/account-service/src/main/java/com/ita/challenges/account/infrastructure/config/SecurityConfig.java
```diff
@@ -20,14 +20,14 @@ public SecurityFilterChain filterChain(HttpSecurity http) throws Exception {
                         .anyRequest().authenticated()
                 )
                 .oauth2Login(oauth2 -> oauth2
-                        .loginPage("/api/account/oauth2/authorization/github")
+                        .loginPage("/auth")
```
kenan-it-academy, line 23: we can't have *anything* outside the `/api/account` prefix... It will simply break when we deploy if we do...

kenan-it-academy (changes requested):
This change can't go live. We need to talk

backend/account-service/src/main/java/com/ita/challenges/account/infrastructure/config/SecurityConfig.java
kenan-it-academy: wait, I misunderstood

kenan-it-academy: Oh dammit, I haven't looked at FRONTEND restrictions, I'm going to need to get back to this

kenan-it-academy (approved):

kenan-it-academy:
Disregard my previous comments :rofl:

Issue #545, opened 2026-05-06
[CHALLENGE] - Load challenge detail dynamically from route param


JordiMiravet:
### Objective
Replace the hardcoded mock data in `ChallengeDetailPage` with dynamic data loading based on the route `id` param.

### Acceptance criteria
- [x] `ChallengeDetailPage` reads the `id` from `ActivatedRoute`
- [x] Uses `ChallengeService.getById(id)` to fetch the challenge
- [x] Challenge data is stored in a `signal<IChallenge | undefined>`
- [x] Template uses `challenge()` signal syntax with optional chaining
- [x] Tests cover: component creation, successful load, and undefined/not-found case
- [x] App compiles with no errors

## Prerequisites
- #544  merged

> These criteria may be refined as implementation evolves.

Issue #540, opened 2026-05-06
[CHALLENGE] - Add submission creation method to API service


JordiMiravet:
### Objective
Enable users to submit a solution (code) to a challenge by sending it to the backend.

### Acceptance Criteria

- [x]  A method exists in the API service to submit a solution to a challenge
- [x]  The method sends the solution data (e.g. state, challenge reference) correctly
- [x]  User ID is handled (A fallback/hardcoded UUID is temporarily used and documented if no user is found)
- [x]  The service can be used from components
- [x]  The application compiles without errors

### Implementation Notes
This issue is resolved by two linked PRs:

- API Service PR: Implemented the postSolution method in the ChallengeApiService to handle the POST request to the backend.
#675 
- Component Integration PR: Updated the challenge-detail-page with a Reactive Form to capture user input and trigger the API service upon submission.
#684

nereame96:
This issue will have 2 PR

PR #678, opened 2026-05-13
Feat/540 submit challenge solution button

nereame96:
[Task here](https://github.com/IT-Academy-BCN/ita-challenges/issues/540)
**Description**
Implemented the challenge submission form and the "Finalize Challenge" button logic.

**Note on Git History**
This PR currently includes commits from branches #545 and #540-api to ensure full functionality and successful builds. Once those dependencies are merged into develop, this PR will reflect only the UI and component changes.

**Key Changes**

- Added IChallengeSubmission interface.
- Integrated ReactiveFormsModule into ChallengeDetailPage.
- Linked submission logic to ChallengeApiService.
- Added unit tests for the submission flow (Happy Path).

### DIFF
--- a/frontend/src/app/features/challenges/pages/challenge-detail-page/challenge-detail-page.html
+++ b/frontend/src/app/features/challenges/pages/challenge-detail-page/challenge-detail-page.html
@@ -2,13 +2,16 @@
 
   <h1>{{ challenge()?.title }}</h1>
   <p>{{ challenge()?.description}}</p>
-  <textarea
-      id="challengeAnswer"
-      rows="10"
-      placeholder="Escriu aquí la resposta del repte...">
-    </textarea>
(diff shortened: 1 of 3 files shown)


andyrouge81 (changes requested):
The scope is clear and the tests cover the main happy path well. I only noticed a few minor details:

* The hardcoded `userId` fallback is probably a temporary measure until the authentication flow is more integrated in future sprints, but it could be worth controlling or documenting it to avoid confusion.
* There are also a few missing `;` in `ChallengeDetailPage` (`inject(...)`, form config, etc.).
* I also noticed that the backend submission endpoint already exists. The only thing is that frontend is currently using `/submissions/finalize`, so I imagine `finalize` may be a temporary mock or provisional adaptation until everything is fully aligned with backend.

Other than that, the PR looks well aligned with the proposed objective.

JordiMiravet (approved):
Everything looks good. Nice job, Nerea!

PR #697, opened 2026-05-18
Fix/security config nginx

eariasvalor:
Fix CORS configuration by centralizing it in Nginx gateway

Remove @CrossOrigin annotations from service controllers and configure CORS at the Nginx gateway level. This ensures a single entry point handles cross-origin requests, avoids duplicated configuration across services, and aligns with the existing network security model where only the gateway is publicly exposed.


soyjuandelgado (approved):
LGTM!

yvangabrieli (approved):
good Idea to ensures a single entry point handles cross-origin requests. Makes it easier. GJ

andyrouge81 (approved):
Nice improvement centralizing CORS in the `Nginx` gateway and removing the duplicated `@CrossOrigin` annotations from the controllers. This makes sense if the gateway is the single public entry point.

I get the frontend is already using this prefix, then everything looks aligned to me. GJ!

JordiMiravet (approved):
Looks good to me! good job Esther!

kenan-it-academy (approved):

PR #797, opened 2026-05-21
feat: Configure logout endpoint to invalidate session and delete cookie.

### LINKED ISSUE #766: [AUTH] - Make logout work properly when refreshing the page
Currently, when a user logs out and refreshes the page, 
they are automatically logged back in.

This task fixes that behavior by properly clearing the session 
when the user logs out, so the app behaves as expected.

Expected behavior:
- After logging out, refreshing the page keeps the user logged out.

YoussefEG09:
Currently, when a user logs out and refreshes the page, 
the session is automatically restored due to the JSESSIONID cookie.

Changes:
- Added logout configuration in SecurityConfig with endpoint POST /api/account/auth/logout
- Invalidate HTTP session on logout
- Delete JSESSIONID cookie on logout
- Added logout endpoint to permitted URLs
- Updated SecurityConfigTest with missing mocks

Expected behavior:
After logout, refreshing the page keeps the user logged out.

Closes #766

### DIFF
--- a/backend/account-service/src/test/java/com/ita/challenges/account/infrastructure/config/SecurityConfigTest.java
+++ b/backend/account-service/src/test/java/com/ita/challenges/account/infrastructure/config/SecurityConfigTest.java
@@ -26,22 +26,22 @@ import static org.assertj.core.api.Assertions.assertThat;
 class SecurityConfigTest {
 
 @Autowired
-    private MockMvc mockMvc;
+private MockMvc mockMvc;
 
 @Autowired
-    private SecurityFilterChain securityFilterChain;
+private SecurityFilterChain securityFilterChain;
 
 @MockBean
-    private ClientRegistrationRepository clientRegistrationRepository;
+private ClientRegistrationRepository clientRegistrationRepository;
 
 @MockBean
-    private TicketRepository ticketRepository;
+private TicketRepository ticketRepository;
 
 @MockBean
-    private UserRepository userRepository;
+private UserRepository userRepository;
 
 @MockBean
-    private RestClient restClient;
+private RestClient restClient;
 
     @Test
     @DisplayName("Should assure the SecurityFilterChain is loaded")
--- a/backend/account-service/src/test/java/com/ita/challenges/account/infrastructure/config/SecurityConfigTest.java
+++ b/backend/account-service/src/test/java/com/ita/challenges/account/infrastructure/config/SecurityConfigTest.java
@@ -1,6 +1,7 @@
 package com.ita.challenges.account.infrastructure.config;
 
 import com.ita.challenges.account.domain.port.out.TicketRepository;
+import com.ita.challenges.account.domain.port.out.UserRepository;
 import org.junit.jupiter.api.DisplayName;
 import org.junit.jupiter.api.Test;
 import org.junit.jupiter.params.ParameterizedTest;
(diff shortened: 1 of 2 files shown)


AlbertMedina (changes requested):
Hey Youssef! I was testing the changes locally, but the original issue is still happening. When I log out and refresh the page, the session is automatically restored and I'm still logged in.

Apart from that, just one change requested in the new test (`shouldPermitLogoutEndpoint`): since you configured a custom `logoutSuccessHandler` to return a `200` status, I think we should check for that explicitly.
Instead of asserting that it's not Forbidden, we can assert that it is actually successful (using `.andExpect(status().isOk());`).
This makes the test much cleaner and more precise. 

Let's double-check the session issue first, and once both things are sorted out, we'll be good to go! Thanks!

YoussefEG09:
Hey Albert! Thanks for testing it.

I've made two changes:

1. Updated the logout configuration to also clear the authentication 
   and added the remember-me cookie deletion, which should fix 
   the session restore issue on refresh.

2. Updated the test to assert status 200 explicitly as you suggested.

Could you test it again and let me know if the issue is resolved? Thanks!

AlbertMedina:
Hey Youssef, thanks for the updates!
The test looks perfect now, thank you! However, I just checked the logout again and the issue is still happening. The cookie is still there, so when I refresh the page, I am automatically logged in again.
To reproduce it, you can start the project using Docker to test this flow locally on your machine. Don't hesitate to ask if you need anything!

JEspinosa1770 (approved):
This endpoint is required to log the user out.

Without this, the user would never log out.

These changes, along with the changes made to the front end, will ensure proper session closure.

**Good work.**

AlbertMedina (approved):
I see that, together with the parallel frontend PR, the logout issue will be resolved. With this in mind, approved from my side. LGTM!

backend/account-service/src/test/java/com/ita/challenges/account/infrastructure/config/SecurityConfigTest.java
```diff
@@ -13,54 +14,69 @@
...
 import static org.springframework.test.web.servlet.result.MockMvcResultMatchers.*;
 import static org.assertj.core.api.Assertions.assertThat;
 
 @WebMvcTest
 @Import(SecurityConfig.class)
 class SecurityConfigTest {
 
- @Autowired
- private MockMvc mockMvc;
+    @Autowired
```
kenan-it-academy: Careful with these random indentation changes, turn them back. This is not part of the PR and it obscures your actual changes. Will review once they're fixed.

kenan-it-academy (changes requested):
Fix formatting errors so it can be reviewed

backend/account-service/src/test/java/com/ita/challenges/account/infrastructure/config/SecurityConfigTest.java
YoussefEG09: Done!! Thank you for you comment anr review!

backend/account-service/src/test/java/com/ita/challenges/account/infrastructure/config/SecurityConfigTest.java
```diff
@@ -13,54 +14,69 @@
...
 import static org.springframework.test.web.servlet.result.MockMvcResultMatchers.*;
 import static org.assertj.core.api.Assertions.assertThat;
 
 @WebMvcTest
 @Import(SecurityConfig.class)
 class SecurityConfigTest {
 
- @Autowired
- private MockMvc mockMvc;
+@Autowired
```
kenan-it-academy, line 28: This is still happening here. It's REALLY hard to keep track of what changes or real like this...

PR #798, opened 2026-05-21
feat(Ticketing): implement addComment and setStatus with unit tests

### LINKED ISSUE #793: [TICKETING] Add ability to comment on tickets: Domain method for comments
**What?**
Add a simple way for mentors to leave comments on tickets. A comment is just a short text saved inside the ticket.

**Why?**
This helps keep mentor notes and feedback connected to the ticket. For now, the goal is only to make tickets able to store a comment in a simple way.

**Behaviour expected:**
Mentors can add a short text comment to a ticket, and the comment is stored inside the ticket in memory.

This is the first step to achieving that, by giving the ticket object a simple way to handle comments. Further steps will make sure users can actually add and see these comments.

yvangabrieli:
Ticket comments 

This PR introduces initial domain support for adding comments to a ticket.

A new addComment() method has been added to the Ticket aggregate as the first step toward allowing mentors to store feedback inside tickets.

The method currently returns this as a temporary placeholder because the Ticket aggregate does not yet include comment storage fields.
The goal of this PR is to introduce the domain behavior first and prepare the integration flow incrementally until the related model changes are merged.

### DIFF
--- a/backend/account-service/src/test/java/com/ita/challenges/account/domain/model/TicketTest.java
+++ b/backend/account-service/src/test/java/com/ita/challenges/account/domain/model/TicketTest.java
@@ -21,15 +21,17 @@ class TicketTest {
         Ticket ticket = Ticket.create("user-1", "Login issue", "desc");
         Ticket updated = ticket.withUpdates(null, "Mentor comment");
         assertEquals("Mentor comment", updated.getComment());
-        assertEquals(ticket.getStatus(), updated.getStatus()); // unchanged
+        assertEquals(ticket.getStatus(), updated.getStatus());
     }
+    
     @Test
     void withUpdates_should_update_only_status() {
         Ticket ticket = Ticket.create("user-1", "Login issue", "desc");
         Ticket updated = ticket.withUpdates(TicketStatus.RESOLVED, null);
         assertEquals(TicketStatus.RESOLVED, updated.getStatus());
-        assertEquals(ticket.getComment(), updated.getComment()); // unchanged (null initially)
+        assertEquals(ticket.getComment(), updated.getComment()); 
     }
+    
     @Test
     void withUpdates_should_update_both_status_and_comment() {
         Ticket ticket = Ticket.create("user-1", "Login issue", "desc");
@@ -37,6 +39,7 @@ class TicketTest {
         assertEquals(TicketStatus.IN_PROGRESS, updated.getStatus());
         assertEquals("Checking this", updated.getComment());
     }
+    
     @Test
     void withUpdates_should_not_change_anything_if_both_params_null() {
         Ticket ticket = Ticket.create("user-1", "Login issue", "desc");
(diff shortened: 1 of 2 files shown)


AlbertMedina (changes requested):
Hi Yvan! Good first step toward adding comments!
However, returning `this` right now is a bit risky because the method fails silently (it looks like it works, but the comment is lost). Instead, I recommend throwing an `UnsupportedOperationException` to make it clear that the feature is not yet implemented. 

Also, I suggest adding a quick unit test to verify this behavior. This ensures that anyone trying to use this method will be explicitly warned until the actual logic is ready.

What do you think? Thanks for pushing this forward!

yvangabrieli:
Good suggestion Albert. All changes have been made and the test as been implemented on TicketTest.

AlbertMedina (approved):
Nice work Yvan! Looks perfect to me now! Thanks for adding the test and updating the exception. Approved!

yvangabrieli:
This PR extends the initial ticket comment support by adding immutable operations for comments and status updates in the Ticket model.

The Ticket now supports:

* `setComment(...)` → adds  mentor comment
* `setStatus(...)` → updates the ticket status

Only one comment is stored per ticket at any time (no comment history). Each update simply overwrites the previous value.

Both methods return a new `Ticket` instance following the immutable domain design.

This PR also renames `addComment` to `setComment` to better reflect the current behaviour.

Unit tests cover comment overwrite, status updates, and chained operations.

Urian1983 (approved):

yvangabrieli:
From the last Commit (24c15462c7ca70fa18b10a34400199fef38c30dc) This method covers all combinations:
Only comment updates
Only status updates
Both at once
Neither—returns a copy with updatedAt changed
You can remove withComment and withStatus as they become redundant.

backend/account-service/src/test/java/com/ita/challenges/account/domain/model/TicketTest.java
```diff
@@ -15,4 +15,34 @@ void should_create_ticket_with_basic_information() {
         assertEquals("Login issue", ticket.getTitle());
         assertEquals("Unable to access my account", ticket.getDescription());
     }
+
+    @Test
+    void withUpdates_should_update_only_comment() {
+        Ticket ticket = Ticket.create("user-1", "Login issue", "desc");
+        Ticket updated = ticket.withUpdates(null, "Mentor comment");
+        assertEquals("Mentor comment", updated.getComment());
+        assertEquals(ticket.getStatus(), updated.getStatus()); // unchanged
```
kenan-it-academy: don't put comments in code, and separate testcases with a newline for readability

kenan-it-academy (changes requested):

backend/account-service/src/test/java/com/ita/challenges/account/domain/model/TicketTest.java
yvangabrieli: My bad. Thank for that.

kenan-it-academy (approved):

PR #806, opened 2026-05-21
feat(api): add find by id endpoint for tickets

### LINKED ISSUE #767: Implement the method to find ticket by id
Implement the controller for FindById: look for what is missing so that this feature works correctly (repository and other classes)

EstherBlacksmith:
Adds GET endpoint to retrieve a specific ticket by identifier with authentication validation. 
Returns 401 if user is not authenticated, 200 with ticket details if found, or 404 if ticket does not exist.

Includes integration test covering the happy path scenario and a mini refactor for unused imports.


AlbertMedina (approved):
Looks good, Esther! Approved from my side!

Urian1983 (approved):
I would add the test for the 401 situation, but since we are doing the happy path  I approve it.

kenan-it-academy:
> I would add the test for the 401 situation, but since we are doing the happy path I approve it.

@Urian1983 this should be added, but in a future PR :wink:

kenan-it-academy (approved):

PR #820, opened 2026-05-22
Updating TicketResponse with new parameters (TicketStatus and Comment)

### LINKED ISSUE #795: [TICKETING] Show comment and current status when viewing a ticket
### What?
Update the ticket response so it includes the current ticket status and the single comment associated with the ticket when a ticket is viewed.

### Why?
To ensure users can see the current state of a ticket together with its latest discussion in a single response, improving consistency between backend data and frontend rendering.

### Behaviour expected
When a ticket is requested, the response payload must include:

1. the current ticket status
2. the single comment associated with the ticket

The response follows the existing Ticket DTO structure and includes standard metadata fields.

Urian1983:
A small update really, it only adds to TicketResponse the parameters TicketStatus and the Comment. Nothing More

### DIFF
--- a/backend/account-service/src/main/java/com/ita/challenges/account/infrastructure/adapter/web/in/TicketController.java
+++ b/backend/account-service/src/main/java/com/ita/challenges/account/infrastructure/adapter/web/in/TicketController.java
@@ -104,6 +104,7 @@ public class TicketController {
         if (!ticket.getUserId().equals(user.getAttribute("login"))) {
             throw new ResponseStatusException(HttpStatus.FORBIDDEN);
         }
+        
 
         return ResponseEntity.ok(new TicketResponse(
                 ticket.getComment(),
--- a/backend/account-service/src/main/java/com/ita/challenges/account/infrastructure/adapter/web/in/TicketController.java
+++ b/backend/account-service/src/main/java/com/ita/challenges/account/infrastructure/adapter/web/in/TicketController.java
@@ -59,6 +59,7 @@ public class TicketController {
         return ResponseEntity.ok(tickets);
     }
 
+
     @PutMapping("/{id}")
     public ResponseEntity<TicketResponse> update(
             @PathVariable String id,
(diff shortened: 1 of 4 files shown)


Spook242 (changes requested):
Hi Josep...!!!

Thank you so much for leaving the comment clarifying the status of the ticket, it helps a lot when reviewing...!!! I've been looking at the code, and here are the reasons why the checks are failing for you:

1. Compilation Error (Build bootJars):
As you rightly mentioned, the TicketResponse class and the Enum aren't finished in develop yet. By trying to use a constructor that doesn't exist, the server can't compile the Java project and fails the check. In these cases, so your PR isn't blocked, I recommend temporarily returning hardcoded dummy data or nulls so it compiles. We can connect it later once our teammates push their parts.

2. Endpoint Route:
On line 96, you declared an empty @GetMapping. You'll need to specify the route to prevent it from conflicting with other endpoints in the TicketController when the app tries to start.

3. Merge Conflicts and Draft PR:
You have some conflicts with develop. I suggest doing a git pull origin develop on your branch to resolve them. Also, since the PR is waiting on other pieces, you can use GitHub's "Convert to draft" option. This way, the team knows it's a Work In Progress and it won't be merged by accident.

Great job...!!! If you need help resolving the conflicts, give me a shout and we can look at it together. 💪

jrffruit518-enrich (changes requested):
Hi Josep 👋

Just a quick clarification on one point so we are aligned with the current state of develop:

I’ve already merged the updated TicketResponse model (with TicketStatus and comment) into develop earlier today, so that part should now be available there. It might be worth pulling the latest changes to double-check.

Also, regarding the PR, since this issue is focused on updating the DTO response (TicketResponse), it might be better to keep the scope aligned with that and avoid introducing additional endpoints in this PR. If the GET endpoint is needed, it could be handled in a separate PR so we keep responsibilities clearer and easier to review.

Let me know if you want to go through it together 👍

Issue #734, opened 2026-05-19
[AUTH] - Make the logout button fully log out the user


JEspinosa1770:
[PR HERE](https://github.com/IT-Academy-BCN/ita-challenges/pull/825) 

Expected behavior:

- Clicking logout calls POST /api/account/auth/logout
- After logout, the user is no longer authenticated
- Refreshing the page does not restore the session

PR #825, opened 2026-05-22
Feat/734 auth   make the logout button fully log out the user

JEspinosa1770:
**Description**

This PR resolves the incomplete logout behavior when clicking the logout button ([[#734](https://github.com/IT-Academy-BCN/ita-challenges/issues/734)]). The AuthService.logout() method now performs a real HTTP request to the backend and clears the user state reactively, and the corresponding tests have been updated to reflect this behavior.

**Changes**

Service: AuthService.logout() now performs a POST request to the logout endpoint and clears the user state reactively via tap(), returning Observable<void>.
Component: LogoutButton.logout() updated to subscribe to the observable and handle errors.
Tests: Updated specs for both AuthService and LogoutButton to reflect the new async behavior.

**Verification Results**

AuthService unit tests passed successfully, covering:

- User signal cleared after successful logout response
- HTTP POST request made to /api/account/auth/logout

LogoutButton component tests passed successfully, covering:

- Button click triggers logout() method
- AuthService.logout() is called

Closes [[#734](https://github.com/IT-Academy-BCN/ita-challenges/issues/734)]

### DIFF
--- a/frontend/src/app/shared/components/logout-button/logout-button.ts
+++ b/frontend/src/app/shared/components/logout-button/logout-button.ts
@@ -12,7 +12,11 @@ export class LogoutButton {
   private authService = inject(AuthService);
 
   logout(): void {
-    this.authService.logout();
+    this.authService.logout().subscribe({
+      next: () => { },
+      error: (error) => {
+        console.error('Error crítico durante el proceso de logout:', error);
+      }
+    });
   }
-
 }
--- a/frontend/src/app/shared/components/logout-button/logout-button.ts
+++ b/frontend/src/app/shared/components/logout-button/logout-button.ts
@@ -14,7 +14,7 @@ export class LogoutButton {
   logout(): void {
     this.authService.logout().subscribe({
       error: (error) => {
-        console.error('Error durante el proceso de logout:', error);
+        console.error('Error durant el procés de logout:', error);
       }
     });
   }
--- a/frontend/src/app/shared/components/logout-button/logout-button.ts
+++ b/frontend/src/app/shared/components/logout-button/logout-button.ts
@@ -9,13 +9,12 @@ import { AuthService } from '../../../features/auth/data-access/auth-service';
 })
 export class LogoutButton {
 
-  private authService = inject(AuthService);
+  private readonly authService = inject(AuthService);
 
   logout(): void {
     this.authService.logout().subscribe({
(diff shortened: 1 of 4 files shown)


frontend/src/app/shared/components/logout-button/logout-button.ts
```diff
@@ -12,7 +12,11 @@ export class LogoutButton {
   private authService = inject(AuthService);
 
   logout(): void {
-    this.authService.logout();
+    this.authService.logout().subscribe({
+      next: () => { },
```
JordiMiravet: `next: () => {}` can probably be removed since it’s currently a no-op.

frontend/src/app/shared/components/logout-button/logout-button.ts
```diff
@@ -12,7 +12,11 @@ export class LogoutButton {
...
+      error: (error) => {
+        console.error('Error crítico durante el proceso de logout:', error);
```
JordiMiravet: This may just be a style preference, but the `console.error` message feels a bit strong ("Critical error..."). A simpler log message might be enough here.

JordiMiravet (changes requested):
Overall, this looks good: moving the logout flow to the backend makes sense, and the subscription in the component is now necessary. Also, the logout now performs an actual logout action instead of only clearing the local user signal.

A couple of minor observations:

JordiMiravet (approved):
Looks good to me now, good job @JEspinosa1770!

frontend/src/app/shared/components/logout-button/logout-button.ts
JEspinosa1770: It's an option I considered, and I decided to keep the callback empty to preserve the complete structure of the subscribe function, but it's true that it doesn't have any function and, therefore, can be considered redundant code. I'll take your idea, @JordiMiravet .

frontend/src/app/shared/components/logout-button/logout-button.ts
JEspinosa1770: The initial message was too strong. I've toned it down a bit and translated it into Catalan.

nereame96 (approved):
Good job Jordi! Looks perfect for me, you properly applied the previous requested changes!

kenan-it-academy (approved):

PR #834, opened 2026-05-23
Feat (challenges) Implement create submitted (#783)

### LINKED ISSUE #783: Implement CreateSubmitted
Objective
Develop the core business logic that triggers when a student decides to finalize and turn in their challenge solution.

Technical Tasks
- Implement the `CreateSubmitted(submission)` method in the domain layer.

Acceptance Criteria
- The method correctly saves the submission state without data loss.
- Unit test for the repository created and passing successfully (`SubmissionRepositoryTest`).

Rafadicandia:
## Objective

Implement the `CreateSubmitted` method to store the details of the challenge submitted by the student in an in-memory repository.

## Technical Tasks

* Implement the `CreateSubmitted(submission)` method in the domain layer.

## Acceptance Criteria

* [ ] The method correctly saves the submission state without data loss.
* [ ] Unit tests for the repository are created and passing successfully (`SubmissionRepositoryTest`).


yvangabrieli (changes requested):
Nice work overall — the tests are clean and the intent of the factories is easy to follow 👍

One thing that stood out to me though is that both factory methods currently duplicate almost the entire object creation logic and only differ by the status value. It may become harder to maintain later if more fields are added, so maybe it would be worth centralizing the shared creation logic into a single internal method.

Spook242 (commented):
Nice work, Rafa...!!! The static factory `createSubmitted` is clean, correctly assigns all timestamps and statuses, and the test coverage is solid. Good job on refactoring the test class variables too.

However, reviewing this alongside your Use Case PR (#839), I noticed a missing piece of domain behavior. 

Right now, we have static factories to create *brand new* submissions (`createInProgress` and `createSubmitted`). But if a student already has a draft (`IN_PROGRESS`), the `Submission` entity lacks an instance method to transition its state.

**Suggestion:**
To fully support the draft workflow in your Use Case, you should add an instance method to this entity (e.g., `public void finalizeSubmission(String finalCode)`). This method should update the `code`, change the `status` to `SUBMITTED`, and refresh the `updatedAt` timestamp.

That way, your Use Case handler can simply load the existing draft and call `submission.finalizeSubmission(command.code())` instead of trying to generate a duplicate entity from scratch.

Could you add that state-transition method here? Otherwise, the rest looks great...!!!

Rafadicandia:
@yvangabrieli I'm 100% agree with you. I implemented this logic for the CreateSubmitted core logic to maintain the previous one, simply because refactoring is out of scope at this point in the development. Therefore, for future Sprints, I think it will be necessary to unify and refactor, but not for now! Thanks!

AlbertMedina (approved):
Nice work, Rafa! Looks good to me!

Rafadicandia:
@Spook242 Thanks for the insight! I've addressed your concern in PR #839. 
Since there is already a repository method to check if a User has already done a final submission, I used this method as a gateway, and if there is a draft, the use case just updates the info and maintains the same SubmissionId so that there are no duplicates. 

Well spotted!

Rafadicandia:
out of scope

yvangabrieli (approved):
Thank you for your feedback. Make sense to me now. It is good that it say documented so in the future it can be implemented and simplified. GJ Rafa

kenan-it-academy (approved):

PR #841, opened 2026-05-24
feat: Get user role by username.

### LINKED ISSUE #730: [AUTH] - Get the user's role

YoussefEG09:
Added a GET endpoint to retrieve the username and role of a user by their GitHub username.
This allows the frontend to display the user's role in the header.

Changes:
- Added GET /api/account/users/{username} endpoint in UserController
- Returns username and role in the response
- Returns 404 if the user is not found
- Added unit tests for both cases

Closes #730

### DIFF
--- a/backend/account-service/src/main/java/com/ita/challenges/account/infrastructure/adapter/web/in/UserController.java
+++ b/backend/account-service/src/main/java/com/ita/challenges/account/infrastructure/adapter/web/in/UserController.java
@@ -1,7 +1,7 @@
 package com.ita.challenges.account.infrastructure.adapter.web.in;
 
 import com.ita.challenges.account.domain.port.out.UserRepository;
-import com.ita.challenges.account.infrastructure.adapter.web.in.dto.UserResponse;
+import com.ita.challenges.account.infrastructure.adapter.web.in.dto.UserRoleResponse;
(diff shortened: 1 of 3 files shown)


JEspinosa1770 (changes requested):
Given that my Java knowledge is rather limited, I wanted to suggest something that might be helpful, if you think it's necessary.

It involves adding a condition (I believe it's done with `@PreAuthorize`) so that only the user who owns the requested role, or an admin, can obtain it. Otherwise, anyone could potentially obtain the role of any user.

I'm just mentioning it as an idea. Let me know if it's worthwhile or unnecessary. I'll review it again based on your feedback.

YoussefEG09:
Thanks for the suggestion, Jordi! After checking with the team lead, 
we decided to keep it simple for now since this endpoint is only used to display the role in the HUD. 
We can add access control later if needed.

JEspinosa1770 (approved):
Ok. 
LGTM.
**Approved.**

AlbertMedina (approved):
Nice work Youssef! Approved!

Urian1983 (approved):
Simple, direct and efficient. I like it

JordiMiravet (approved):
LGTM! Good job Youssef

kenan-it-academy (approved):

Issue #768, opened 2026-05-20
Fix creation of the ticket, adding missing methods and changing the hardcode


EstherBlacksmith:
Correct the creation of a ticket and add what is missing: check if any more needs to be done or if the entire cycle is complete

PR #851, opened 2026-05-24
updating Ticket Controller to adapt it to TicketStatus and comment

Urian1983:
# User Story

As a developer,  
I want to update the `TicketController` to align with the latest domain changes,  
so that the API correctly handles the new `TicketStatus` and `Comment` fields introduced in the `Ticket` domain model.

## Changes Included

- Updated `TicketController` methods to support the new domain structure.
- Adapted the corresponding Records/DTOs to reflect the addition of:
  - `TicketStatus`
  - `Comment`
- Fixed the incomplete `@PostMapping` (`create`) implementation found in the first commit of the branch, which was causing test failures.
- Added the missing call to the `save` method in `TicketRepository` from `TicketController`.

## Notes

No modifications were made to:
- `TicketRepository`
- `InMemoryTicketRepository`
- Existing tests in `TicketControllerRepository`

The goal of these changes is to restore consistency between the controller layer and the updated domain model while preserving the existing repository implementation and test coverage.

### DIFF
--- a/backend/account-service/src/main/java/com/ita/challenges/account/infrastructure/adapter/web/in/TicketController.java
+++ b/backend/account-service/src/main/java/com/ita/challenges/account/infrastructure/adapter/web/in/TicketController.java
@@ -32,13 +32,16 @@ public class TicketController {
 
         String userId = user != null ? user.getAttribute("login") : "temp-user";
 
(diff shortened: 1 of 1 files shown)


Urian1983:
#Duplicated of #768

PR #875, opened 2026-05-27
731 auth   display the role in the header on the right

### LINKED ISSUE #731: [AUTH] - Display the role in the header, on the right
Role comes from back

Expected behavior:

- The active user's role is displayed on the right side of the header

JEspinosa1770:
### Show the logged-in user's role in the header

Description:
When a user is logged in, their role is now displayed in the header of the application alongside their username and avatar. If the role is not available, a default 'CONVIDAT' label is shown as fallback. This way, users can easily see at a glance which account and role they are currently using, improving the overall experience and making the interface feel more informative.

To ensure the user data is available on any route and not only when navigating through the profile page, the header now fetches the user on initialization. Additionally, the role fetch in AuthService has been made fault-tolerant: if the role endpoint is unavailable or returns an error, the user is still loaded correctly without a role, preventing unwanted redirects to the login page.

### DIFF
--- a/frontend/src/app/features/auth/data-access/auth-service.spec.ts
+++ b/frontend/src/app/features/auth/data-access/auth-service.spec.ts
@@ -6,6 +6,7 @@ import { vi } from 'vitest';
 
 import { AuthService } from './auth-service';
 import { AuthUser } from '../models/auth-user.model';
+import { Role } from '../../../core/models/role.enum';
 
 const MOCK_USER: AuthUser = {
   username: 'mockUser',
@@ -67,17 +68,6 @@ describe('AuthService', () => {
     vi.unstubAllGlobals();
   });
 
-  it('should fetch user from API', async () => {
-    const promise = firstValueFrom(service.fetchUser());
-
-    const req = httpMock.expectOne('/api/account/auth/me');
-    expect(req.request.method).toBe('GET');
-    req.flush(MOCK_USER);
-
-    const result = await promise;
-    expect(result).toEqual(MOCK_USER);
-  });
-
   it('should clear current user on logout', async () => {
     service.setUser(MOCK_USER);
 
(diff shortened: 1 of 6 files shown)


frontend/src/app/layout/components/header/header.html
```diff
@@ -2,7 +2,17 @@
...
-    <h4 >
+    @if (user()?.role) {
+      <h4 class="user__profile">
+        {{ user()?.role }}
+      </h4>
+    } @else {
+       <h4 class="user__profile">
+        CONVIDAT
+      </h4>
+    }
```
JordiMiravet: The logic is correct, but the role rendering in `header.html` could probably be simplified by replacing the conditional block with something like:

```html
<h4 class="user__profile">
   {{ user()?.role ?? 'CONVIDAT' }}
</h4>
```

JordiMiravet (changes requested):
Nice improvement overall 👍
The only thing I’d change:

frontend/src/app/layout/components/header/header.html
JEspinosa1770: You're absolutely right. Much more concise and elegant. Thank you.

JordiMiravet (approved):
Much cleaner now, great job Jordi

nereame96 (approved):
Looks good to me! Well done implementing the previous suggestions!

frontend/src/app/features/auth/data-access/auth-service.spec.ts
```diff
@@ -91,4 +81,19 @@ describe('AuthService', () => {
...
   });
+
+  it('should fetch user and role from API', async () => {
+    const promise = firstValueFrom(service.fetchUser());
+    httpMock.expectOne('/api/account/auth/me').flush(MOCK_USER);
+    httpMock.expectOne(`/api/account/users/${MOCK_USER.username}/role`).flush({ role: Role.GUEST });
+    expect(await promise).toEqual({ ...MOCK_USER, role: Role.GUEST });
+  });
+
+  it('should load user without role if role endpoint fails', async () => {
```
kenan-it-academy: fallback behavior should be a separate PR

kenan-it-academy (changes requested):

frontend/src/app/features/auth/data-access/auth-service.spec.ts
JEspinosa1770: Understood! I have removed the fallback behavior test and the corresponding catchError from AuthService. I will address that in a separate PR once this one is approved.

kenan-it-academy (approved):

PR #881, opened 2026-05-29
feat(auth-service): handle role endpoint failure #731

### LINKED ISSUE #731: [AUTH] - Display the role in the header, on the right

JEspinosa1770:
**Description:**

When fetching the logged-in user, AuthService also requests their role from a separate endpoint. If that endpoint is unavailable or returns an error, the observable would fail, causing the auth guard to redirect the user back to the login page in an uncontrolled way.

This PR adds a catchError to the role fetch so that, if it fails, the user is still loaded correctly without a role instead of breaking the authentication flow. The fallback sets the role to undefined, which the header already handles by displaying the default 'CONVIDAT' label.


AlbertMedina (approved):
Looks good to me, no issues found. Approved!

JordiMiravet (approved):
Looks good to me. Good job Jordi.

kenan-it-academy (approved):

PR #883, opened 2026-06-01
731 auth   display the role in the header on the right

JEspinosa1770:
## Description
When a user logs in with GitHub, they may not yet exist in the database (no role assigned). The previous implementation returned a 404 in that case, which caused the frontend auth flow to fail and redirect the user back to the login page.

## Changes
- `SecurityConfig`: allow unauthenticated access to `/api/account/users/**`
- `UserController`: return `GUEST` role instead of 404 when user is not found in the repository
- update UserControllerTest to expect GUEST role when user not found

## Related issue
[#731](https://github.com/IT-Academy-BCN/ita-challenges/issues/731)


JordiMiravet (approved):
LGTM!

EstherBlacksmith (approved):

kenan-it-academy (approved):

PR #919, opened 2026-06-03
"refactor: remove default anonymous user"

### LINKED ISSUE #936: [AUTH] Remove default user fetching and update UI components
Update the Angular application so it stops making HTTP requests to the removed backend endpoint and handles the unauthenticated state properly.

**Tasks:**
- [x] Remove `fetchUser()` from `AuthService`.
- [x] Simplify `authGuard` to block access if no user is in memory.
- [x] Update `ProfilePageComponent` to remove the fallback HTTP subscription.
- [x] Update `Header` component to rely only on the computed user signal.

Spook242:
If a person is not logged in, the system no longer returns a default or dummy user profile.

Closes #936

### DIFF
--- a/backend/account-service/src/main/java/com/ita/challenges/account/infrastructure/adapter/web/in/AuthController.java
+++ b/backend/account-service/src/main/java/com/ita/challenges/account/infrastructure/adapter/web/in/AuthController.java
@@ -61,23 +61,4 @@ public class AuthController {
         );
     }
 
-    @GetMapping("/auth/me")
-    public ResponseEntity<AuthUserDto> authMe(@AuthenticationPrincipal OAuth2User user) {
-        return ResponseEntity.ok(mapUser(user));
-    }
-
-    private AuthUserDto mapUser(OAuth2User user) {
-        if (user == null) {
-            return new AuthUserDto("anonymous", null);
-        }
-
-        String login = user.getAttribute("login");
-        String avatarUrl = user.getAttribute("avatar_url");
-
-        if (login == null) {
-            login = "unknown";
-        }
-
-        return new AuthUserDto(login, avatarUrl);
-    }
 }
--- a/backend/account-service/src/main/java/com/ita/challenges/account/infrastructure/adapter/web/in/AuthController.java
+++ b/backend/account-service/src/main/java/com/ita/challenges/account/infrastructure/adapter/web/in/AuthController.java
@@ -61,4 +61,26 @@ public class AuthController {
         );
     }
 
-}
+    @GetMapping("/auth/me")
+    public ResponseEntity<AuthUserDto> authMe(@AuthenticationPrincipal OAuth2User user) {
+        if (user == null) {
+            return ResponseEntity.status(HttpStatus.UNAUTHORIZED).build();
+        }
+
+        return ResponseEntity.ok(mapUser(user));
+
+    }
+
+        private AuthUserDto mapUser (OAuth2User user){
+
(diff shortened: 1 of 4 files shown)


Urian1983 (approved):

Rafadicandia (changes requested):
Hi Pedro! I'm not quite sure if deleting the whole endpoint is what is needed here. Shouldn't it be better if we just ensure that if the user is not logged in, we just show a message and banner the entrance?
I mean, maybe it's better to change the if sentence into a ¨User not logged in¨ and stop his entrance there. 
By the way, there is another endpoint that assigns the Gest Role when not logged in, maybe that endpoints needs  a fix and stop assigning that role and just throw an exception error. 

Tell me what you think!

Spook242:
> Hi Pedro! I'm not quite sure if deleting the whole endpoint is what is needed here. Shouldn't it be better if we just ensure that if the user is not logged in, we just show a message and banner the entrance? I mean, maybe it's better to change the if sentence into a ¨User not logged in¨ and stop his entrance there. By the way, there is another endpoint that assigns the Gest Role when not logged in, maybe that endpoints needs a fix and stop assigning that role and just throw an exception error.
> 
> Tell me what you think!

Hi Rafa! You absolutely hit the nail on the head.

I just confirmed this exact behavior with Luis. He clarified that if a user hasn't explicitly logged in, we shouldn't be assigning a GUEST session or any default user behind the scenes. Since the ticket is strictly about removing the default user, returning Role.GUEST falls exactly into the anti-pattern we are trying to clean up.

I've just updated the UserController to throw a 404 Not Found exception instead of returning a default role. I'll push the commit right now. Thanks a lot for catching that

JEspinosa1770 (commented):
Hi Pedro,

This is a comment to help me understand these changes.
The change in UserController seems correct to me. It now does what it's supposed to do.

But I don't understand the deletion of the auth/me endpoint. I've tested it locally and it returns a 404 error because it can't find auth/me. Perhaps this pull request depends on changes made in another one? Or maybe I'm missing something.

If you could clarify this for me, I would appreciate it.

Rafadicandia (changes requested):
Hi, Pedro! Please reevaluate if deleting the endpoint ¨auth/me¨ is needed here. I think it's not accurate for the task

Spook242:
I have restored /auth/me so we don't break the current frontend. However, please note that we recently created a new and cleaner endpoint (/api/account/users/me) in PR #938. Once the frontend is updated to use the new one, we can safely delete this old /auth/me in a future refactor

YoussefEG09 (changes requested):
Looks good Pedro! The changes are clean and well focused.

Could you add/update the tests to cover the new behavior?
Specifically:
- authMe should return 401 when the user is not authenticated
- getUserRole should return 404 when the user is not found

Spook242:
Hi Youssef...!!! Thanks for catching this. You were completely right, the test suite needed to reflect the new behavior now that the default dummy user is gone.

I've just updated both AuthControllerTest and UserControllerTest.

The changes are pushed and the tests are green! Let me know if everything looks good to you now.

JEspinosa1770 (approved):
LGTM. There are no more login errors. 

Good decision.

Rafadicandia (approved):
Hi, @Spook242 Good job refactoring! Now it looks like a very straight forward way to validate and entrace and throw the propper error responde. 
LGTM!

kenan-it-academy (approved):

YoussefEG09 (approved):

Issue #896, opened 2026-06-02
[AUTH] Remove default user and handle unauthorized access


vicenlu:
### Description
The goal of this issue is to clean up how the platform handles unauthenticated visitors. Currently, the system provides a "default user" profile when someone accesses the app without logging in. We need to stop this behavior so the system correctly identifies when no one is logged in. 

To fix this, we will remove the old backend endpoint that provides this default user, and we will also remove the frontend code that asks for it.

### Acceptance Criteria (DoD)
- [x] If a person is not logged in, the system no longer returns a default or dummy user profile.
- [x] The backend endpoint responsible for handling this default user behavior is completely removed from the code.
- [x] The frontend (Angular) is updated so it no longer makes requests to the removed backend endpoint.

PR #925, opened 2026-06-03
Feature/896 frontend cleanup

Spook242:
### Description
This Pull Request updates the frontend (Angular) to stop asking the backend for a "default" or "anonymous" user when a visitor is not logged in. This is part of the cleanup for Issue #896 to ensure the system correctly identifies unauthenticated visitors.

### Acceptance Criteria (DoD)
- The frontend application no longer makes HTTP requests to the backend looking for a default user profile.
- The `AuthService`, Navigation Header, Profile page, and Auth Guard have been simplified and updated to stop using the removed feature.
- The application compiles successfully without errors.

### DIFF
--- a/backend/account-service/src/main/java/com/ita/challenges/account/infrastructure/adapter/web/in/AuthController.java
+++ b/backend/account-service/src/main/java/com/ita/challenges/account/infrastructure/adapter/web/in/AuthController.java
@@ -61,23 +61,4 @@ public class AuthController {
         );
     }
 
-    @GetMapping("/auth/me")
-    public ResponseEntity<AuthUserDto> authMe(@AuthenticationPrincipal OAuth2User user) {
-        return ResponseEntity.ok(mapUser(user));
-    }
-
-    private AuthUserDto mapUser(OAuth2User user) {
-        if (user == null) {
-            return new AuthUserDto("anonymous", null);
-        }
-
-        String login = user.getAttribute("login");
-        String avatarUrl = user.getAttribute("avatar_url");
-
-        if (login == null) {
-            login = "unknown";
-        }
-
-        return new AuthUserDto(login, avatarUrl);
-    }
 }
--- a/backend/account-service/src/main/java/com/ita/challenges/account/infrastructure/adapter/web/in/AuthController.java
+++ b/backend/account-service/src/main/java/com/ita/challenges/account/infrastructure/adapter/web/in/AuthController.java
@@ -29,6 +29,27 @@ public class AuthController {
         this.restClient = restClient;
     }
 
+    @GetMapping("/auth/me")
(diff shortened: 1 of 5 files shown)


Urian1983 (approved):

JEspinosa1770 (changes requested):
Hi Pedro,

I've been reviewing the proposed changes in this PR and I think there might be a problem. The header.ts file was listening until a user was active. By removing that, no user data is displayed in the header, because even if a user logs in later, the header signal isn't updated.

Also, the deleted fetchUser() method in auth-service.ts was the one that requested user data from the backend. By removing it, there's nothing from the front end requesting data from the backend, so no user data is received on the front end.

If the goal of the task is for the front end not to obtain any user data, as the sub-issue description seems to indicate, the changes made are correct, but I don't think that's what task 896 requires. If I'm wrong, please let me know, but I don't think the expected result is being achieved.

Spook242:
Hi Jordi, thanks for the detailed review...!!! You are absolutely right. I got a bit carried away cleaning up the mock user data and accidentally removed the actual data-fetching mechanism. Luis told me to delete it and I got carried away. 😄 😄 😄 

I need to keep fetchUser() alive because it's exactly what will call the new GET /me backend endpoint I implemented in PR #944. I will restore the header.ts listener and the service method in a new commit right away so the header updates correctly. Good catch...!!!

JEspinosa1770 (approved):
Hi Pedro,

With the current code, I don't see the bugs that occurred with the previous version being possible.
Good work.

backend/account-service/src/main/java/com/ita/challenges/account/infrastructure/adapter/web/in/AuthController.java
```diff
@@ -61,23 +61,4 @@ public ResponseEntity<UserResponse> create(@RequestBody UserRequest request) {
         );
     }
 
-    @GetMapping("/auth/me")
```
kenan-it-academy, line 64: So we're removing this? I'm very confused cause a different PR is adding `/me`.

kenan-it-academy (changes requested):

backend/account-service/src/main/java/com/ita/challenges/account/infrastructure/adapter/web/in/AuthController.java
Spook242: Luis told me to delete the backend endpoint and I got a bit carried away 😄  Then I realized it, and I did it in the GitHub profile and avatar PR. I'm fixing it right now.

Rafadicandia (changes requested):
Hi, @Spook242 I see a potencial issue regarding the urls. I mentioned it in PR #951. Can you correct the Url as they where before? That way we dont's brake de gateway routing.

Spook242:
Unnecessary PR

PR #943, opened 2026-06-04
Persist challenge deletion in database

### LINKED ISSUE #933: Remove any challenge
We want to remove any challenges perpetually from the database.

Urian1983:
This change adds the delete method in JsonFileChallengeRepository class, and the happy past test to its test file. Thanks to it now the deletion of the challenges should persist since it now deletes from the generated file and not from RAM memory-

### DIFF
--- a/backend/challenge-service/src/main/java/com/itachallenges/challengeservice/catalog/infrastructure/adapter/web/out/persistence/JsonFileChallengeRepository.java
+++ b/backend/challenge-service/src/main/java/com/itachallenges/challengeservice/catalog/infrastructure/adapter/web/out/persistence/JsonFileChallengeRepository.java
@@ -73,7 +73,7 @@ public class JsonFileChallengeRepository implements ChallengeRepository {
         try {
             persistToFile();
         } catch (IOException e) {
-            throw new RuntimeException("Failed to persist after delete");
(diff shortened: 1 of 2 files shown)


cristhianchulca49 (approved):
"Simple and functional! We can now remove the `InMemorySubmissionRepository` class

Spook242 (changes requested):
Great job with the deletion logic, Josep...!!! Just a minor best practice detail: in the catch block, when throwing the RuntimeException, you forgot to pass the original exception e as the cause. If we leave it like this and the file writing fails, we will lose the stack trace of the actual error, making debugging very difficult.

The test for the happy path is super clear, awesome...!!!
As a suggestion to make the PR coverage a perfect 10/10: since you added the defensive logic to throw a NoSuchElementException if the ID doesn't exist in the storage map, it would be great to add a test with a fake ID using assertThrows.

Let me know what you think. Great work...!!!

Urian1983:
@Spook242 

I added your proposed changes in the second commit.

Spook242 (approved):
Awesome...!!! Thank you for implementing those changes so quickly, Josep. The exception handling is now perfectly robust, and the test coverage is 10/10. Great job...!!!

kenan-it-academy (approved):

Issue #915, opened 2026-06-03
[CHALLENGE] Improve the display of the Challenges list


JEspinosa1770:
### Description:

Improve the challenge list by creating a card for each challenge, allowing users to easily view and differentiate its details. The required data to be displayed are the Title, Description, Difficulty, and Language. Each card includes the predefined buttons.

Essential: the description must be a maximum of three lines

### Out of scope: 

Detailed editing in Figma's design style, pagination, and filters.

### PRs
[948](https://github.com/IT-Academy-BCN/ita-challenges/pull/948)
[949](https://github.com/IT-Academy-BCN/ita-challenges/pull/949)
[950](https://github.com/IT-Academy-BCN/ita-challenges/pull/950)

JEspinosa1770:
The possibility of unifying styles across the different challenge pages was discussed at the meeting on 09/06/26, and it was agreed that this would be a backlog task if there is not enough time to implement it in this sprint.

PR #949, opened 2026-06-05
feat(challenges): improve challenge card UI with icons and style structure #915

JEspinosa1770:
Changes included:

- Add Tabler Icons CDN link to index.html for icon support
- Add getDifficultyClass method to dynamically assign difficulty tag styles
- Update challenge card template with language badge, hardcoded tags, 
  difficulty pill and stats row (rocket, heart, date)

Not included: css file

### DIFF
--- a/frontend/src/app/features/challenges/pages/challenges-list-page/challenges-list-page.spec.ts
+++ b/frontend/src/app/features/challenges/pages/challenges-list-page/challenges-list-page.spec.ts
@@ -106,20 +106,4 @@ describe('ChallengesListPage', () => {
     expect(component.challenges().length).toBe(initialLength - 1);
     expect(component.challenges().some(c => c.id === CHALLENGES_MOCK[0].id)).toBe(false);
   });
-
-  it('should return easy class for EASY difficulty', () => {
-    expect(component.getDifficultyClass('EASY')).toBe('tag tag--difficulty-easy');
-  });
-
-  it('should return medium class for MEDIUM difficulty', () => {
-    expect(component.getDifficultyClass('MEDIUM')).toBe('tag tag--difficulty-medium');
-  });
-
-  it('should return hard class for HARD difficulty', () => {
-    expect(component.getDifficultyClass('HARD')).toBe('tag tag--difficulty-hard');
-  });
-
-  it('should return default tag class when difficulty is undefined', () => {
-    expect(component.getDifficultyClass(undefined)).toBe('tag');
-  });
 });
--- a/frontend/src/app/features/challenges/pages/challenges-list-page/challenges-list-page.spec.ts
+++ b/frontend/src/app/features/challenges/pages/challenges-list-page/challenges-list-page.spec.ts
@@ -15,7 +15,7 @@ describe('ChallengesListPage', () => {
   let mockChallengeService: any;
 
   beforeEach(async () => {
-    mockChallengeService = { 
+    mockChallengeService = {
       loadAll: vi.fn().mockReturnValue(of(CHALLENGES_MOCK)),
       delete: vi.fn().mockReturnValue(of(void 0)),
     };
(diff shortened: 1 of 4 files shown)


Spook242 (approved):
Awesome work putting the HTML structure together to match the CSS grid, Jordi...!!! The implementation of the getDifficultyClass method using template literals is very clean, and I really appreciate that you took the time to update the unit tests to cover the new logic. That's top-notch...!!!

I just have two suggestions:

1. Hardcoded Data (Tags and Stats): > I noticed that the tags (Lògica, Variables, Fonaments) and the stats (rocket, heart, date) are currently hardcoded in the HTML template. This means every single challenge card will display the exact same tags and stats. If we don't have this data coming from the backend Challenge model yet, that's completely fine for a UI layout PR...!!! However, I highly recommend adding an HTML comment (e.g., ``) just so we don't forget to refactor it later.

2. Angular Performance Pro-tip:
We recently mentioned this in another PR, but it's a great practice for the whole team: calling component methods like getDifficultyClass() and getDifficultyLabel() directly from the template inside a @for loop can impact performance. Angular's Change Detection runs these functions on every cycle. For future refactors, consider using Angular Pipes for this kind of data transformation, or map these CSS classes and labels directly to the Challenge objects in the TypeScript file before passing them to the template.

Great job overall...!!!! Let me know what you think about adding the TODO note for the hardcoded values.

JEspinosa1770:
Thanks for your observations, Pedro.

As soon as I can, I'll add some comments to clarify that the hardcoded code will need to be changed in the future, once we have all the data.

Regarding your comment on the Angular Performance Pro-tip, thanks for the advice! In this case, the challenge list is a signal, so change detection only runs when an update is made, minimizing the performance impact. That said, we agree it's good practice, and we can refactor to Pipe in a future iteration to maintain consistency across the entire codebase.

isaacmg-bit (approved):
All good! The hardcoded stuff is the only way right now, while we wait for the backend development.

alexgesti (approved):
All good from my side! The hardcoded data makes sense for now since we're still waiting on the backend to provide the real structure.

frontend/src/app/features/challenges/pages/challenges-list-page/challenges-list-page.spec.ts
```diff
@@ -106,4 +106,20 @@ describe('ChallengesListPage', () => {
     expect(component.challenges().length).toBe(initialLength - 1);
     expect(component.challenges().some(c => c.id === CHALLENGES_MOCK[0].id)).toBe(false);
   });
+
+  it('should return easy class for EASY difficulty', () => {
+    expect(component.getDifficultyClass('EASY')).toBe('tag tag--difficulty-easy');
```
kenan-it-academy: this feels very brittle with the use of `.toBe` as it will break as soon as any other class is added for any reason. I also am unsure if this is a good test in general, as those classes are only aesthetic and don't change functionality 🤔

kenan-it-academy (commented):
I'm unsure of some things here with the tests...

frontend/src/app/features/challenges/pages/challenges-list-page/challenges-list-page.spec.ts
JEspinosa1770: What you say about the tests not adding much makes sense, since the logic is only for visual styling; there's no business logic involved.

Therefore, if you're okay with it, I'll remove them, @kenan-it-academy

kenan-it-academy (approved):

frontend/src/app/features/challenges/pages/challenges-list-page/challenges-list-page.spec.ts
kenan-it-academy: I think it makes sense in this case

PR #951, opened 2026-06-05
"feat: redirect to frontend after successful oauth2 login"

### LINKED ISSUE #924: [AUTH] Redirect Users After Successful Authentication
## Description

Automatically redirect authenticated users to the main page.

## Technical Tasks

- Handle successful authentication callback.
- Implement navigation to the main page.
- Validate redirect flow after login.

## Acceptance Criteria

- Users are redirected to the main page after successful authentication.
- Redirect works for both manual login and restored sessions.
- No navigation errors occur during the authentication process.

### Out of Scope

End-to-end validation of the GitHub OAuth flow depends on backend authentication configuration and callback handling.

Spook242:
## Description
This PR resolves the issue where users were not being redirected back to the Angular application after a successful GitHub OAuth2 login. Previously, the backend was routing authenticated users to a default `/profile` endpoint, causing a blank page or 404 error.

## Changes Made
* Updated `SecurityConfig.java` in the `oauth2Login` filter chain.
* Configured `.defaultSuccessUrl("http://localhost:4200/", true)` to ensure Spring Security automatically pushes the user back to the Frontend application once GitHub authentication is complete.

## Important Note for Frontend Team (@alexgesti / @Isaac / @Jordi)
To make this full flow work, the login buttons in the Angular application must bypass Angular's internal router and make a direct request to the backend. 
Please ensure the login buttons (in the `Header` and the `Auth` page) are updated from `<button>` to standard `<a>` tags pointing to the backend port (e.g., 8080):
`<a href="http://localhost:8080/api/account/oauth2/authorization/github" class="your-css-classes">Log In</a>`

## Related Issue
Closes #924


Urian1983 (approved):
You should talk with the Frontend team since the changes affects a lot of issues that are being working now in different parts of the project.

Rafadicandia (changes requested):
Hi, Pedro! I see you are trying clean the login flow, but I think this approach is colidding with the gateway we are implementing. I don't think hardcodig local adresses is a good idea, since the gateway manages every route, back and front. Weshould keep the security config as it is or use real frontend directions. 
Can you have a look at this?

Thanks!!

Spook242:
Hi Rafa...!!! You are absolutely right. Hardcoding localhost:4200 would break the flow once deployed and bypass the gateway routing. I have refactored the SecurityConfig to use a dynamic @Value("${app.frontend.url}") property instead. This way, we can inject the correct Gateway/Frontend URL depending on the environment (local vs production) without altering the code. Pushing the fix right now, thanks for catching it...!!!

Rafadicandia:
@Spook242 There are still hardcoded URL at front end. O the other hand, the security config I think it should stay as it was and if needed, change de login destination page for a frontend route. The cors are not needed here, since it's being handled by the gateway.

JEspinosa1770 (commented):
There are changes to the front end that the Auth team will need to adapt; however, these changes shouldn't include a hardcoded URL. Since nginx is being used, the front end can use a relative path and define the target where necessary so that nginx can handle the redirection correctly.

Similarly, hardcoded URLs in auth-service.ts are also not a good practice. It's always better to use relative paths and a definition in the environment, so if modifications are needed, they only need to be made in one place and not in different parts of the code.
It would be advisable to fix these two points, if not in this pull request, then in a later one.

Regarding the back end code, everything looks fine, but I lack the context to ensure it fulfills the expected functionality. Your back end team will be able to advise you better than I can on this point.

Rafadicandia (approved):
Good work, Pedro. I think nonw is more accurate and within scope of the usecase.

LGTM!

kenan-it-academy (approved):

PR #953, opened 2026-06-05
feat(challenge): auto-populate editForm with challenge data on init #942

### LINKED ISSUE #942: [CHALLENGE] Connect Form Controls to Challenge Data
### Description

Implement the logic to auto-populate the reactive form (editForm) when the user enters the editing screen. The task involves capturing the active challenge details and injecting them into the corresponding form controls (id, title, description, solution, difficulty, and language) so the form loads ready for modification.

### Out of scope

Handling data persistence/saving back to the server, advanced caching, or building complex offline rollback mechanics.

### PR
[953](https://github.com/IT-Academy-BCN/ita-challenges/pull/948)

JEspinosa1770:
Implement auto-population of the reactive form when the user enters the edit screen.

- Inject ActivatedRoute to capture the challenge id from the route params
- Add ngOnInit to fetch the challenge data from the backend via getById()
- Use patchValue to populate the form fields (id, title, description, solution, difficulty, language) with the retrieved data
- Handle the undefined case from the observable to ensure type safety
- Added tests to verify functionality with onInit

### DIFF
--- a/frontend/src/app/features/challenges/challenges.routes.ts
+++ b/frontend/src/app/features/challenges/challenges.routes.ts
@@ -12,7 +12,7 @@ export const CHALLENGES_ROUTES: Routes = [
     component: ChallengesListPage,
   },
   {
-    path: 'edit',
+    path: 'edit/:id',
     component: EditChallengePage,
   },
   {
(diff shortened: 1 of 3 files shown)


isaacmg-bit (approved):
Good!

There are subscribes and nothing destroys those subscriptions, but it should not be an issue as I've followed the getById to its source (challengeservice -> challenge api service) and I see it's just an HTTP petition.

In any case I'm not super sure if it would be a good practice to use something to destroy the subscription, but I don't think it is necessary!

Good work Jordi!

Rafadicandia (approved):
Hi, Jordi! Good job, clean, nice error handling and good test coverage. 

LGTM

kenan-it-academy (approved):
Approving but this challenge page is getting big enough we should start thinking about splitting it

Issue #893, opened 2026-06-02
[TICKET] Display error feedback on ticket creation failure


vicenlu:
As a platform user, I need to receive clear visual feedback when my ticket submission fails, so I can understand that the action was not completed and my request was not saved. This feedback should be immediate, noticeable, and help me determine whether I need to retry the action or take other steps to resolve the issue. The system should clearly distinguish between successful and failed submissions to prevent uncertainty about whether my ticket was properly recorded.

PR #954, opened 2026-06-07
Feat [TICKET] display error feedback on ticket creation failure #893

Rafadicandia:
## What was done
- Added error message display to the ticket creation form
- Connected the form to the backend error response (400 Bad Request with message)
- Extracted the error message from the backend response in TicketApiService
- Forced Angular change detection so the message appears immediately on first submit
- Added an automated test that simulates a failed ticket creation and verifies the error message is shown
## Acceptance Criteria
✅ Basic user-facing error messaging is in place
✅ Create ticket page is connected to the error system
✅ Users see a clear, friendly message when ticket creation fails
✅ Automated test covers the failure scenario and asserts the message appears in the DOM

### DIFF
--- a/frontend/src/app/features/tickets/pages/create-ticket-page/create-ticket-page.ts
+++ b/frontend/src/app/features/tickets/pages/create-ticket-page/create-ticket-page.ts
@@ -1,5 +1,5 @@
-import { Component, inject } from '@angular/core';
-import { FormBuilder, ReactiveFormsModule } from '@angular/forms';
+import { Component, inject, ChangeDetectorRef } from '@angular/core';
+import { FormBuilder, ReactiveFormsModule, Validators } from '@angular/forms';
 import { TicketApiService } from '../../data-access/ticket-api.service';
 import { ITicketRequest } from '../../models/iticket-request.interface';
 import { Router } from '@angular/router';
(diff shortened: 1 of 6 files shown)


JEspinosa1770 (approved):
Rafa, great front-end code!

Error handling is important because it gives the user the necessary information in case of any failure. 

**Good work!**

isaacmg-bit (approved):
Good placeholder solution until we get to implement toast notifications!

Another thing we should have present is that the hardcoded CSS stuff should be replaced for the styles.css variables as soon as we do the create ticket page layout. I think we should do a backlog task so we do not forget to do it!

Good work!

frontend/src/app/features/tickets/pages/create-ticket-page/create-ticket-page.ts
```diff
@@ -11,20 +11,29 @@ import { Router } from '@angular/router';
...
 
   onSubmit() {
+    this.errorMessage = null;
+
     const newTicket = this.ticketForm.value as ITicketRequest
 
     this.ticketService.create(newTicket).subscribe({
-      next: () => {this.goTickets();}
+      next: () => {this.goTickets();},
+      error: (err) => {
```
kenan-it-academy, line 33: We need a test for this in the spec file too

kenan-it-academy (changes requested):
Missing a test, everything else looks good!

frontend/src/app/features/tickets/pages/create-ticket-page/create-ticket-page.ts
Rafadicandia: @kenan-it-academy Done!

kenan-it-academy (approved):

PR #956, opened 2026-06-07
FEAT [TICKET]: return error message body on invalid create ticket request

Rafadicandia:
## Summary

Modified createTicket in TicketController to return the validation error message in the response body when the request contains blank title or description fields.
Changed the return type from ResponseEntity<TicketResponse> to ResponseEntity<?> to allow returning a plain text error message.

## Test coverage:
Added/updated test should_return_400_when_sending_blank_fields_in_create_ticket in TicketControllerTest to verify that:

When a TicketRequest with blank title and description is sent to POST /api/account/tickets
The endpoint returns HTTP 400 Bad Request status
The response body contains the error message "request vacia"

### DIFF
--- a/backend/account-service/src/main/java/com/ita/challenges/account/infrastructure/adapter/web/in/TicketController.java
+++ b/backend/account-service/src/main/java/com/ita/challenges/account/infrastructure/adapter/web/in/TicketController.java
@@ -33,12 +33,15 @@ public class TicketController {
     }
 
     @PostMapping
-    public ResponseEntity<TicketResponse> createTicket(
+    public ResponseEntity<?> createTicket(
             @RequestBody TicketRequest request,
             @AuthenticationPrincipal OAuth2User user) {
 
         if (user == null || user.getAttribute("login") == null) {
             return ResponseEntity.<TicketResponse>status(HttpStatus.UNAUTHORIZED).build();
+        }
+        if (request.title().isBlank() || request.description().isBlank()){
+            return ResponseEntity.status(HttpStatus.BAD_REQUEST).body("No empty fields allowed");
         }
 
         String userId = user.getAttribute("login");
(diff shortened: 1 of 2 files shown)


YoussefEG09 (approved):
LGTM! Works correctly for the current MVP scope.

Small note for the future: ResponseEntity<?> loses type safety, 
could be improved with a dedicated error DTO when we revisit this.

Rafadicandia:
@YoussefEG09 We can review this totally! I found this the fastest solution to return different ResponseEntity accordinng to a dto, or a error with body. Since we are not, yet, handling custom error nither a propper GlobalExceptionHandler, I thought this may be the easyest approach. Thanks!! I'll will add the GlobalExceptionHandler to tech debt

JEspinosa1770 (approved):
Hi Rafa,

Simple, elegant, and atomized.

Good job.

Spook242 (changes requested):
Hi Rafa...!!!

I've been reviewing the changes, and I think we can tweak them a bit to align more with Spring Boot best practices. Here are a couple of suggestions:

Manual validation vs. Annotations: Instead of checking manually with if (ticketRequest.title().isBlank()), it would be better to use Jakarta validations.

API Contract: By doing the manual check, you had to change the method signature to ResponseEntity<?> in order to return a String. Ideally, we shouldn't break the API contract; we should always return a JSON (even for errors). By using Jakarta validations, Spring will throw an exception that we can cleanly catch to return a proper response.

Tests and messages: In the test, you check status().isBadRequest() perfectly, but we'd need to add an assertion to check the response body and verify that the correct error text is being returned.

Let me know what you think!

Rafadicandia:
@Spook242 I agree with you completely, and as I mentioned to @YoussefEG09 since we dont have a proper GlobalHandlerException we just throw a JSON with generic info that builds Spring Boot. I prefered this not so beatiful way and added the GlobalExeptionHander to tech debt since it will be needed for future Exeption casting messages.
In the other hand, thanks for your suggestion about the test, I applyed them and now its much more safe this way!

Thanks you for your review!

Spook242 (approved):
Hi Rafa, I hadn't thought about the GlobalExceptionHandler. Also, the test looks perfect now with the new content assertion...!!! Great job...!!!

cristhianchulca49 (approved):
Great work on this implementation. The logic is clear, and you have successfully implemented the necessary authorization check and basic input validation to keep the demo stable.

kenan-it-academy (approved):

PR #977, opened 2026-06-11
enabling official response from backend

### LINKED ISSUE #976: [CHALLENGE] Bonus: wiring the submission frontend with the backend
In the current iteration of the application, when a submission is sent to get the correct response to a problem, the backend is not the one that gives the response, but the frontend. The objective of this sub-issue is simple: modify the backend and give it the ability to return the solution.

Urian1983:
This PR adds the next changes in the backend:

- Adds a new record/DTO called SubmissionResultResponse that sends the official response to the backend from the frontend.
- A new parameter (boolean revealOfficialResponse) has been added to FinalizeSubmisionRequest, a record/DTO that already existed.
- SubmissionController has been updated with a new version of the finalize method; now instead of returning to the backend a void as a confirmation of the PostMapping, it returns the solution through the new RecordDTO (SubmissionResultResponse).

Since I am working in parallel with my partners, the first commit will be a draft since it is possible that small changes and a new commit would be needed to adjust the files to the new changes. Suggestions and corrections are more than welcomed.

### DIFF
--- a/backend/challenge-service/src/main/java/com/itachallenges/challengeservice/submission/infrastructure/adapter/in/web/controller/SubmissionController.java
+++ b/backend/challenge-service/src/main/java/com/itachallenges/challengeservice/submission/infrastructure/adapter/in/web/controller/SubmissionController.java
@@ -25,9 +25,9 @@ public class SubmissionController {
     private final SubmissionRepository submissionRepository;
     private final ChallengeRepository challengeRepository;
 
-    public SubmissionController(SaveDraftSubmissionUseCase saveDraftSubmissionUseCase, SubmissionRepository submissionrepository, ChallengeRepository challengeRepository) {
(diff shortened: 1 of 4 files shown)


Rafadicandia (approved):
Hi! @Urian1983 Well done! The refactor of the repository is necessary in the entire endpoint, so I find this PR very accurate despite it does a bit more than what the description suggest. 

LGTM!

cristhianchulca49 (approved):
This is a great extra! It helps the frontend track whether a student has submitted their answer or not, which I think adds a lot of value.

YoussefEG09 (changes requested):
LGTM! Good job Josep!! Just a tiny naming nit: "submissionrepository" parameter on SubmissionController constructor should be "submissionRepository" (capital R) to follow Java conventions.

YoussefEG09 (approved):
Very good job Josep!! Approved for me!!!

kenan-it-academy (approved):

PR #1001, opened 2026-06-19
feat(tickets): show backend response feedback on ticket creation and add tests #995

### LINKED ISSUE #995: [TICKET] Show feedback message when a ticket has been created
### Description
Implement user feedback messages in the frontend when a ticket creation action is performed. The application needs to handle different HTTP response codes returned by the backend and display an appropriate, simple message to the user indicating the outcome of the request.

> ⚠️ **Note:** To speed up development, complex styling and layout alignment with the final Figma design are currently excluded from this task. Basic, clear, and functional UI alerts/labels are sufficient.

### Acceptance Criteria
- [ ] **Success - 201:** When filling out the form with valid data and clicking "Create ticket", the message **"Ticket creat correctament"** must be displayed.
- [ ] **Invalid data - 400:** When submitting the form with incorrect or incomplete data, the message **"Dades del ticket incorrectes"** must be displayed.
- [ ] **No permissions - 403:** When attempting to create a ticket with an unauthorized user, the message **"No tens permís per crear aquest ticket"** must be displayed.
- [ ] **Other errors:** Upon any other server failure response (e.g., 500), the message **"Error en crear el ticket"** must be displayed.
- [ ] **Navigation:** The displayed message must automatically disappear if the user navigates away to another section.

### Out of Scope
- Implementing advanced UI/UX layouts, custom animations, or pixel-perfect styling matching the final Figma blueprint.
- Designing or modifying backend API endpoints or status codes.
- Managing advanced session recovery or routing logic for authorization errors (just displaying the error message is enough for this ticket).

JEspinosa1770:
## Description

Add a non-blocking message below the create-ticket form to inform the user how the creation process ended (success, validation error, permission error or unexpected error), based on the HTTP status returned by the backend.

- Replace errorMessage property and ChangeDetectorRef with a ticketMessage signal, removing the need for manual change detection
- Map HTTP status codes to user-facing messages (201, 400, 403, default) 
- Delay navigation to /tickets by 2s on success so the message is readable, clearing the timeout on component destroy
- Add tests covering the success message/navigation flow and the 400 error message

## Outside of this PR

- No tests have been added for the 403 error case or the generic error message; only 201 and 400 tests are covered as a proof of concept for the pattern. These can be added in a follow-up commit/PR using the same approach.
- No visual styling is applied to the message (plain text, as requested for this phase); the final design (colors, icons, position, entry/exit animation, etc.) is left for future tasks.

### DIFF
--- a/frontend/src/app/features/tickets/pages/create-ticket-page/create-ticket-page.ts
+++ b/frontend/src/app/features/tickets/pages/create-ticket-page/create-ticket-page.ts
@@ -1,4 +1,4 @@
-import { Component, inject, ChangeDetectorRef } from '@angular/core';
+import { Component, inject, signal, DestroyRef } from '@angular/core';
 import { FormBuilder, ReactiveFormsModule, Validators } from '@angular/forms';
 import { TicketApiService } from '../../data-access/ticket-api.service';
 import { ITicketRequest } from '../../models/iticket-request.interface';
@@ -11,33 +11,47 @@ import { Router } from '@angular/router';
   styleUrl: './create-ticket-page.css',
 })
 export class CreateTicketPage {
-  private readonly cdr = inject(ChangeDetectorRef);
-  readonly ticketService = inject(TicketApiService)
-  readonly fb = inject(FormBuilder)
-  readonly router = inject(Router)
+  readonly ticketService = inject(TicketApiService);
+  readonly fb = inject(FormBuilder);
+  readonly router = inject(Router);
+  private readonly destroyRef = inject(DestroyRef);
 
-  errorMessage: string | null = null;
+  readonly ticketMessage = signal<string>('');
 
   ticketForm = this.fb.group({
     title: [''],
     description: ['']
-  })
+  });
 
   onSubmit() {
-    this.errorMessage = null;
+    this.ticketMessage.set('');
 
-    const newTicket = this.ticketForm.value as ITicketRequest
+    const newTicket = this.ticketForm.value as ITicketRequest;
 
     this.ticketService.create(newTicket).subscribe({
-      next: () => {this.goTickets();},
+      next: () => {
+        this.ticketMessage.set('Ticket creat correctament');
+        const redirectId = setTimeout(() => this.goTickets(), 2000);
+        this.destroyRef.onDestroy(() => clearTimeout(redirectId));
+      },
       error: (err) => {
-        this.errorMessage = 'Error al crear el ticket: ' + err.error;
-        this.cdr.detectChanges();
+        this.ticketMessage.set(this.getMessageForStatus(err.status));
       }
     });
   }
 
   goTickets() {
     this.router.navigate(['/tickets']);
-  }    
+  }
+
+  private getMessageForStatus(status: number): string {
+    switch (status) {
+      case 400:
+        return 'Dades del ticket incorrectes';
+      case 403:
+        return 'No tens permís per crear aquest tiquet';
+      default:
+        return "Error en crear el ticket";
+    }
+  }
 }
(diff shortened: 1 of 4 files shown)


isaacmg-bit (approved):
Looks good!

mpujazon (approved):
Great work. The implementation is clean and matches the goal: user feedback now reflects backend outcomes clearly. Approved from my side.

ecantosf:
Hi! This is my first review – I know the PR already has the required approvals, but I wanted to take a look as a learning exercise.

I like the signal-based approach and the clean status mapping. Two small things I noticed:

1. The old `errorMessage` block in the HTML seems unused now – maybe it could be removed.
2. Adding `Validators.required` to the form fields might improve UX if the backend expects them.

Overall, the PR looks solid and achieves its goal. Nice work!

JEspinosa1770:
Thanks for your review, @ecantosf.

Just to comment, the new HTML only uses `ticketMessage`; the previous `errorMessage` was removed. This might be a problem with the display of the old and new code.

Regarding the validations, you're right. Adding validations to the form would be very beneficial, but it wasn't part of the tasks assigned to this project. However, it would be a good idea to add it to the backlog to work on in future sprints. Good point.

frontend/src/app/features/tickets/pages/create-ticket-page/create-ticket-page.ts
```diff
@@ -11,33 +11,47 @@ import { Router } from '@angular/router';
...
+    this.ticketMessage.set('');
 
-    const newTicket = this.ticketForm.value as ITicketRequest
+    const newTicket = this.ticketForm.value as ITicketRequest;
 
     this.ticketService.create(newTicket).subscribe({
-      next: () => {this.goTickets();},
+      next: () => {
+        this.ticketMessage.set('Ticket creat correctament');
+        const redirectId = setTimeout(() => this.goTickets(), 2000);
```
kenan-it-academy: Why would we ever delay this? No timeouts. Ever.

frontend/src/app/features/tickets/pages/create-ticket-page/create-ticket-page.ts
```diff
@@ -11,33 +11,47 @@ import { Router } from '@angular/router';
...
-    this.errorMessage = null;
+    this.ticketMessage.set('');
 
-    const newTicket = this.ticketForm.value as ITicketRequest
+    const newTicket = this.ticketForm.value as ITicketRequest;
 
     this.ticketService.create(newTicket).subscribe({
-      next: () => {this.goTickets();},
+      next: () => {
+        this.ticketMessage.set('Ticket creat correctament');
```
kenan-it-academy: The message seems like it should be shown on destination.

kenan-it-academy (changes requested):
This PR should be split in at least two: the success case and the error case. Additionally, we should NEVER introduce artificial delays at any point. If delays are needed for an implementation to work, that means the implementation is wrong 😉

frontend/src/app/features/tickets/pages/create-ticket-page/create-ticket-page.ts
JEspinosa1770: When ticket creation is successful, the user is redirected to /tickets (ticket list). Displaying a success message doesn't make much sense if it won't be seen, and the delay is simply to give the user two seconds to view the message.

It's true that the flow could be changed, so there's no automatic redirection upon success, and the user has to manually enter the URL to return to the /tickets page. 
Alternatively, there could be no success message at all, and only error messages could be handled, with no redirection. 

What do you think would be more appropriate, @kenan-it-academy

JEspinosa1770: Thank you for your feedback. 
Following your instructions, I will now remove the success message.

kenan-it-academy (approved):
I still don't like having the success and error cases together, but I'll let it slide

JEspinosa1770:
> I still don't like having the success and error cases together, but I'll let it slide

No, they're not together anymore. There's no message about success, only about mistakes.

Issue #990, opened 2026-06-18
[CHALLENGE] Filter challenges by language


vicenlu:
<img width="629" height="228" alt="Image" src="https://github.com/user-attachments/assets/c2088765-b1de-44eb-a847-694cea4a0316" />

PR #1019, opened 2026-06-20
[CHALLENGE] Create filter method in the Backend

### LINKED ISSUE #1010: [CHALLENGE] Create filter method in backend
As a student browsing the challenges catalog, I want to narrow the list down to a specific programming language, so I only see challenges I can practice with the language I am learning or working on.

Acceptance Criteria
- When no language filter is applied, the student sees the full list of available challenges (unchanged from today).
- When a language is selected, the student sees only the challenges that belong to that language.
- When the language selection is cleared, the full list is restored.
- If the selected language has no challenges yet, the student sees an empty list with no error — the catalog simply shows there is nothing to practice in that language right now.
- The languages a student can filter by are the ones currently supported by the platform: Java, SQL, Python, JavaScript, PHP, TypeScript. Any other value is treated as an invalid request and returns no data.

ecantosf:
## Summary

Add an optional `language` query parameter to `GET /api/challenge` to filter
the challenges list by programming language. Backend counterpart of parent
issue #990 *[CHALLENGE] Filter challenges by language*.

## Changes

- Extend `ChallengeRepository` port with `findByLanguage(ChallengeLanguage)`.
- Implement the filter in `InMemoryChallengeRepository` (returns an immutable list).
- Add optional `language` query param to `ChallengeController.findAll()`.
  Spring binds the enum and returns 400 on invalid values automatically.
- Add unit tests for the repository (2) and controller (2).
- Drive-by SonarLint cleanups in adjacent code (see commit message for the full list).

## Verification

- `./gradlew :challenge-service:test` from `backend/`: BUILD SUCCESSFUL, 51 tests pass (4 new).
- Manual check with the `full` Docker profile running:
  - `GET /api/challenge` returns the full list.
  - `GET /api/challenge?language=JAVA` returns only Java challenges.
  - `GET /api/challenge?language=COBOL` returns 400.

## Out of scope

- Aligning the catalog context with an `application/` use case layer.
- Filtering by multiple criteria, pagination, sorting.

Closes #1010

### DIFF
--- a/backend/challenge-service/src/main/java/com/itachallenges/challengeservice/catalog/infrastructure/adapter/web/in/controller/ChallengeController.java
+++ b/backend/challenge-service/src/main/java/com/itachallenges/challengeservice/catalog/infrastructure/adapter/web/in/controller/ChallengeController.java
@@ -3,6 +3,7 @@ package com.itachallenges.challengeservice.catalog.infrastructure.adapter.web.in
 import com.itachallenges.challengeservice.catalog.domain.model.Challenge;
 import com.itachallenges.challengeservice.catalog.domain.port.out.ChallengeRepository;
 import com.itachallenges.challengeservice.catalog.domain.valueobject.ChallengeId;
+import com.itachallenges.challengeservice.catalog.domain.valueobject.ChallengeLanguage;
 import com.itachallenges.challengeservice.catalog.infrastructure.adapter.web.in.dto.ChallengeResponse;
 import com.itachallenges.challengeservice.catalog.infrastructure.adapter.web.in.dto.ChallengeRequest;
 import lombok.AllArgsConstructor;
@@ -39,14 +40,24 @@ public class ChallengeController {
     }
 
     @GetMapping
-    public ResponseEntity<List<ChallengeResponse>> findAll() {
-        List<ChallengeResponse> challenges = repository.findAll()
-                .stream()
-                .map(c -> new ChallengeResponse(c.getId().toString(), c.getTitle().toString(), c.getDescription().toString(), c.getLanguage(), c.getDifficulty(), c.getSolution().toString()))
-                .toList();
-        return ResponseEntity.ok(challenges);
-    }
+    public ResponseEntity<List<ChallengeResponse>> findAll(
+            @RequestParam(required = false) ChallengeLanguage language
+    ) {
+        List<Challenge> challenges = (language == null)
+                ? repository.findAll()
+                : repository.findByLanguage(language);
 
+        List<ChallengeResponse> response = challenges.stream()
+                .map(c -> new ChallengeResponse(
+                        c.getId().toString(),
+                        c.getTitle().toString(),
+                        c.getDescription().toString(),
+                        c.getLanguage(),
+                        c.getDifficulty(),
+                        c.getSolution().toString()))
+                .toList();
+        return ResponseEntity.ok(response);
+    }
 
     @GetMapping("/{id}")
     public ResponseEntity<ChallengeResponse> find(@PathVariable String id) {
@@ -68,7 +79,6 @@ public class ChallengeController {
         repository.delete(ChallengeId.of(id));
         return ResponseEntity.noContent().build();
     }
-
 
     @PutMapping("/{id}")
     public ResponseEntity<ChallengeResponse> update(
--- a/backend/challenge-service/src/main/java/com/itachallenges/challengeservice/catalog/infrastructure/adapter/web/out/persistence/InMemoryChallengeRepository.java
+++ b/backend/challenge-service/src/main/java/com/itachallenges/challengeservice/catalog/infrastructure/adapter/web/out/persistence/InMemoryChallengeRepository.java
@@ -3,6 +3,7 @@ package com.itachallenges.challengeservice.catalog.infrastructure.adapter.web.ou
 import com.itachallenges.challengeservice.catalog.domain.model.Challenge;
 import com.itachallenges.challengeservice.catalog.domain.port.out.ChallengeRepository;
 import com.itachallenges.challengeservice.catalog.domain.valueobject.ChallengeId;
+import com.itachallenges.challengeservice.catalog.domain.valueobject.ChallengeLanguage;
 import org.springframework.stereotype.Repository;
 
 import java.io.IOException;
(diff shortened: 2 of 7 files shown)


ecantosf:
@kenan-it-academy, I'd appreciate it if you could take a look when you can – it's the first backend code for Sprint 1.

backend/challenge-service/src/main/java/com/itachallenges/challengeservice/catalog/infrastructure/adapter/web/out/persistence/InMemoryChallengeRepository.java
```diff
@@ -20,19 +21,15 @@ public class InMemoryChallengeRepository implements ChallengeRepository {
 
     @Override
     public Challenge save(Challenge challenge) {
-        try{
-            storage.put(challenge.getId(), challenge);
-            return challenge;
-        }catch(RuntimeException e){ throw new RuntimeException("Error saving challenge" + e.getMessage());}
-
+        storage.put(challenge.getId(), challenge);
```
Urian1983: I don't understand why you have removed the try catch clause

backend/challenge-service/src/main/java/com/itachallenges/challengeservice/catalog/infrastructure/adapter/web/in/controller/ChallengeController.java
```diff
@@ -39,15 +40,25 @@ public ResponseEntity<ChallengeResponse> create(@RequestBody ChallengeRequest re
...
+    public ResponseEntity<List<ChallengeResponse>> findAll(
+            @RequestParam(required = false) ChallengeLanguage language
+    ) {
+        List<Challenge> challenges = (language == null)
+                ? repository.findAll()
+                : repository.findByLanguage(language);
+
+        List<ChallengeResponse> response = challenges.stream()
+                .map(c -> new ChallengeResponse(
+                        c.getId().toString(),
```
Urian1983, line 52: If already some of these variables return String as a type, then the toString could be noise.

Urian1983 (changes requested):

backend/challenge-service/src/main/java/com/itachallenges/challengeservice/catalog/infrastructure/adapter/web/out/persistence/InMemoryChallengeRepository.java
ecantosf: Fair point on procedure — this was a drive-by cleanup triggered by SonarLint (java:S112: avoid throwing generic RuntimeException), but I shouldn't have touched code outside the filter feature without checking with you first. Apologies for that.

The technical reasoning behind the removal: the original try/catch caught RuntimeException only to re-throw a new RuntimeException with the message concatenated, which drops the original stack trace and adds no information the caller can act on. That's the smell Sonar flags.

I've reverted save() to its original state in the latest commit so this PR stays focused on the filter feature. If the cleanup is welcome, I can submit it as a separate refactor PR for proper review.

backend/challenge-service/src/main/java/com/itachallenges/challengeservice/catalog/infrastructure/adapter/web/in/controller/ChallengeController.java
ecantosf: The `.toString()` calls are needed here because the getters return value objects (`ChallengeId`, `ChallengeTitle`, `ChallengeDescription`, `ChallengeSolution`), not `String`. The `ChallengeResponse` constructor expects `String` for those fields, so the conversion is required for the code to compile. The two getters without `.toString()` (`getLanguage`, `getDifficulty`) return enums that the response carries as enums.

This pattern predates this PR — you'll find the same `.toString()` mapping in `create()`, `find()` and `update()` in this controller. I kept it consistent here so we wouldn't end up with mixed styles in the same class.

If we want to drop the `.toString()` boilerplate across the controller, a cleaner option would be to extract a `ChallengeResponseMapper` (or a static `ChallengeResponse.from(Challenge)` factory) and centralize the mapping in one place. Happy to do that as a separate refactor PR if you think it's worth it.

Urian1983 (approved):
LGFM!

isaacmg-bit (approved):
My knowledge in Java is minimal, but this looks clean to me.

isaacmg-bit (approved):

isaacmg-bit (approved):

Spook242 (approved):
Great implementation and test coverage...!!! Great job Eduard...!!! 👍🏻

cristhianchulca49 (approved):
Looks good for me ! great job.

backend/challenge-service/src/test/java/com/itachallenges/challengeservice/catalog/infrastructure/adapter/web/out/persistence/InMemoryChallengeRepositoryTest.java
```diff
@@ -119,6 +114,28 @@ void should_throw_when_deleting_non_existing_challenge() {
                 .isInstanceOf(NoSuchElementException.class)
                 .hasMessageContaining("Challenge not found with id: " + nonExistentId);
     }
-}
 
+    @Test
+    void should_return_only_challenges_with_given_language() {
```
kenan-it-academy: filter tests should probably be in a separate file

backend/challenge-service/src/test/java/com/itachallenges/challengeservice/infrastructure/adapter/web/in/controller/ChallengeControllerTest.java
```diff
@@ -139,4 +138,25 @@ void should_return_200_with_challenge_when_finding_by_id() throws Exception {
                 .andExpect(jsonPath("$.difficulty").value("EASY"))
                 .andExpect(jsonPath("$.solution").value("Test Solution"));
     }
+
+    @Test
+    void should_return_200_with_filtered_list_when_requesting_challenges_by_language() throws Exception {
```
kenan-it-academy: an edge that should be tested (but might be a different PR) is what do we do with a CORRECT filter that returns NO challenges. Is that a 200 with empty set or a 404?

kenan-it-academy (commented):
Two quick comments I want you to see, but might approve after talking about them

backend/challenge-service/src/test/java/com/itachallenges/challengeservice/catalog/infrastructure/adapter/web/out/persistence/InMemoryChallengeRepositoryTest.java
ecantosf: Done. I created InMemoryChallengeRepositoryFilterTest with the two filter tests, and applied the same split at the controller layer (ChallengeControllerFilterTest) for consistency.

backend/challenge-service/src/test/java/com/itachallenges/challengeservice/infrastructure/adapter/web/in/controller/ChallengeControllerTest.java
ecantosf: Test verifies 200 + [] for no-match filters (REST convention: collection exists, query ok). 404 would be wrong; this HTTP test makes the contract explicit, complementing repo coverage.

PR #1027, opened 2026-06-21
feat(auth): assign 'Student' role automatically on OAuth2 login and test

### LINKED ISSUE #1021: [AUTH] Assign the 'Student' role automatically when a user logs in

Spook242:
Currently, when a new user logged in via GitHub OAuth2, they were not explicitly saved in our database with a default role, leading to a 404 error when fetching their role and forcing the frontend to fall back to a "Guest" (CONVIDAT) display.

This PR introduces a custom OAuth2 user service to intercept the login flow, automatically registering new users in the database with the STUDENT role.

- Register new users automatically with the STUDENT role
- Wire the custom service into SecurityConfig
- Implement unit tests for CustomOAuth2UserService logic

Resolves #1021

### DIFF
--- a/backend/account-service/src/main/java/com/ita/challenges/account/infrastructure/config/CustomOAuth2UserService.java
+++ b/backend/account-service/src/main/java/com/ita/challenges/account/infrastructure/config/CustomOAuth2UserService.java
@@ -13,6 +13,7 @@ import org.springframework.stereotype.Service;
 
 import java.util.Collections;
 import java.util.List;
+import java.util.Optional;
 
 @Service
 public class CustomOAuth2UserService extends DefaultOAuth2UserService {
@@ -27,11 +28,7 @@ public class CustomOAuth2UserService extends DefaultOAuth2UserService {
     public OAuth2User loadUser(OAuth2UserRequest userRequest) throws OAuth2AuthenticationException {
         OAuth2User oAuth2User = super.loadUser(userRequest);
 
-        processOAuth2User(oAuth2User);
-
-        List<SimpleGrantedAuthority> authorities = Collections.singletonList(
-                new SimpleGrantedAuthority("ROLE_STUDENT")
-        );
+        List<SimpleGrantedAuthority> authorities = processOAuth2User(oAuth2User);
 
         return new DefaultOAuth2User(
                 authorities,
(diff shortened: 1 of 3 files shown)


mpujazon (approved):
Nice fix to the OAuth2 onboarding flow. Automatically persisting new OAuth users with the STUDENT role and covering it with unit tests is a solid improvement that should eliminate the fallback/404 behavior. Approved!

JEspinosa1770 (approved):
Good work.
It's a good solution for ensuring user role persistence.
Necessary to continue with the project.

Urian1983 (approved):
The role now doesn't return HTTP 404 code and tickets are now unlocked. Very Good Job @Spook242 

I must say that I wanted to solve the same problem, but using JWT instead of Github own methods.

backend/account-service/src/main/java/com/ita/challenges/account/infrastructure/config/CustomOAuth2UserService.java
```diff
@@ -0,0 +1,51 @@
...
+        this.userRepository = userRepository;
+    }
+
+    @Override
+    public OAuth2User loadUser(OAuth2UserRequest userRequest) throws OAuth2AuthenticationException {
+        OAuth2User oAuth2User = super.loadUser(userRequest);
+
+        processOAuth2User(oAuth2User);
+
+        List<SimpleGrantedAuthority> authorities = Collections.singletonList(
```
kenan-it-academy: I feel like this is either unnecessary or should be inside the `if (login != null...` block you put in the `processOauth2User` function

backend/account-service/src/main/java/com/ita/challenges/account/infrastructure/config/CustomOAuth2UserService.java
```diff
@@ -0,0 +1,51 @@
...
+                new SimpleGrantedAuthority("ROLE_STUDENT")
+        );
+
+        return new DefaultOAuth2User(
```
kenan-it-academy, line 33: is there a reason why we can't just `return oAuth2User`?

kenan-it-academy (changes requested):

backend/account-service/src/main/java/com/ita/challenges/account/infrastructure/config/CustomOAuth2UserService.java
Spook242: Thanks Kenan...!!!
I've updated the code. Now, the function that processes the user is fully responsible for checking the database and returning their correct roles, whether it's a new user or an existing one. Let me know how it looks now.

backend/account-service/src/main/java/com/ita/challenges/account/infrastructure/config/CustomOAuth2UserService.java
Spook242: I looked into returning just oAuth2User, but doing so drops the ROLE_STUDENT from our database, and Spring Security won't recognize the permissions. However, I've pushed a refactor (addressing your other comment) that cleans up the flow and properly injects the authorities into the DefaultOAuth2User. Let me know if this new structure looks better...!!!

kenan-it-academy (approved):

PR #1028, opened 2026-06-21
[CHALLENGE] Create and add filter dropdown button

alexgesti:
## Description

The "Filtros" button has been positioned according to the Figma design and triggers the dropdown interaction.

Inside the dropdown menu, three difficulty options are displayed (Fácil, Media, Difícil). These are currently static and will be used in a future PR for actual filtering logic.

---

## What’s done

- Created filter dropdown component
- Added toggle open/close behavior
- Implemented overlay to close dropdown when clicking outside
- Rendered static difficulty options inside the dropdown
- Integrated dropdown into challenges list page
- Added unit tests for:
  - Component creation
  - Dropdown rendering
  - Open/close behavior
  - Overlay interaction


isaacmg-bit (changes requested):
Alex,

You need to make the PR way smaller :(

I would do, minimum, three to four PRs:

- Dropdown css, dropdown ts, dropdown html
- dropdown tests
- challenge list html, ts and tests
- challenge list css

That should make every PR no more than 100 lines.

And about the buttons, I would do a single button component and inject it into the dropdown. That way you can still cut more lines from the dropdown component and the single button component would have its own PR... and the app would be more modular.

PR #1035, opened 2026-06-22
[CHALLENGE] Create dropdown menu and button

alexgesti:
## Description

A new dropdown component has been created for the challenges list page filter UI.

The component includes the basic dropdown behavior:
- opens when the trigger button is clicked
- closes when clicking the trigger again
- closes when clicking outside through the overlay

This PR only covers the creation of the dropdown component and its interaction behavior. Its content and filtering logic will be handled in future PRs.

---

## What’s done

- Created the filter dropdown component
- Added toggle open/close behavior
- Added overlay to close the dropdown when clicking outside

PR #1036, opened 2026-06-22
[CHALLENGE] Testing filter dropdown & icon add

alexgesti:
## Description

This PR adds test coverage for the dropdown component and includes the filter icon asset used in the dropdown trigger button.

It only covers the testing setup for the component and the static asset addition. Any further UI or filtering behavior will be handled in future PRs.

---

## What’s done

- Added unit tests for the dropdown component
- Added the filter icon asset used by the dropdown trigger


JEspinosa1770 (approved):
Great job with the Dropdown tests! You've covered all possible flows (click, double-click, overlay) and the logic is perfect.

As a suggestion: you're currently searching for elements using CSS classes (.dropdown__menu). If we change styles or migrate to a library like Tailwind in the future, these tests will break even if the component works. To avoid this, Angular allows us to add the `data-testid` attribute to the HTML and use `By.css('[data-testid="..."]')`. This protects the tests against any design changes.

Another tip: in cumulative pull requests, it's helpful to specify the approval order. This one contains code from version 1035, so it might be good practice to specify that version 1035 should be merged first, and then this one.

isaacmg-bit (changes requested):
As with the other PR, please, use the variables existing in the styles.css file.

Plus, I would say this PR is still kind of big :(

alexgesti:
All changes from this file will appear on https://github.com/IT-Academy-BCN/ita-challenges/pull/1037 as requested.

alexgesti:
PR decrapted, need to accept https://github.com/IT-Academy-BCN/ita-challenges/pull/1035 and merge before continue.

PR #1037, opened 2026-06-22
[CHALLENGE] Display dropdown on challenges list

alexgesti:
## Description

The filter dropdown has been integrated into the challenges list page.

The dropdown component is now part of the page layout and renders correctly within the header actions section. The existing tests have been updated to reflect this integration and verify that the dropdown trigger is present in the challenges list view.

---

## What’s done

- Integrated dropdown component into challenges list page
- Updated existing tests to reflect component integration
- Added test coverage to ensure the dropdown trigger is rendered in the challenges list page


JEspinosa1770 (approved):
Great final work on the difficulty dropdown.

A quick suggestion: Would you mind renaming it to DifficultyDropdown? Since it's customized for difficulty levels, this way we can keep the generic name Dropdown free in case we need to create a reusable base component for other filters in the app later.
It's an idea, if not for now, then for making the change later.

isaacmg-bit (changes requested):
As with the other two PR's, please, use variables from the styles.css

Plus this PR is still too big!

alexgesti:
> Great final work on the difficulty dropdown.
> 
> A quick suggestion: Would you mind renaming it to DifficultyDropdown? Since it's customized for difficulty levels, this way we can keep the generic name Dropdown free in case we need to create a reusable base component for other filters in the app later. It's an idea, if not for now, then for making the change later.

The idea is to use this one for all filters, as you can see for the example, all filters are inside the same dropdown, I will change the name to "dropdown-filters" just as a general way of using it for the filters.

image

Urian1983:
WTF, man ... try to divide in in three PR at least, it is very long.

JEspinosa1770 (changes requested):
I've been reviewing the files for this second pull request and I've noticed that the files from the first generic dropdown are still appearing in the project—the CSS, TS, and HTML files. When I created the new ones, that code ended up duplicated in the project.

To avoid dragging dead code into the develop branch, could you please delete the files from the old dropdown component in this same branch and push the commit? This will clean up the project and close the first pull request directly, and significantly reduce the number of lines in this pull request.

alexgesti:
PR decrapted, need to accept https://github.com/IT-Academy-BCN/ita-challenges/pull/1035 and merge before continue.

PR #1039, opened 2026-06-22
[TICKET] get the list of mentors

### LINKED ISSUE #1033: [TICKET] Get the list of Mentors
* **Goal:** Let the system look up and display all mentors.
* **Acceptance Criteria:**
  * The system can fetch and show a list containing *only* users who are registered as Mentors.
  * Each mentor in the list is clearly identified by their username.

cristhianchulca49:
### Description
This PR exposes a new API endpoint to retrieve the list of available mentors. It allows the system to look up and display mentors for future assignments.

### Technical Changes
* Created a new `GET "/mentors"` route in the controller layer.
* Implemented filtering logic to query the `InMemoryUserRepository` and return **only** users with the `Role.MENTOR` value.
* Used the `userName` as the unique identifier in the response payload to align with current infrastructure logic.

### DIFF
--- a/backend/account-service/src/main/java/com/ita/challenges/account/domain/port/out/UserRepository.java
+++ b/backend/account-service/src/main/java/com/ita/challenges/account/domain/port/out/UserRepository.java
@@ -2,10 +2,13 @@ package com.ita.challenges.account.domain.port.out;
 
 import com.ita.challenges.account.domain.model.User;
 
+import java.util.List;
 import java.util.Optional;
 
 public interface UserRepository {
     Optional<User> findByUsername(String username);
 
     User save(User user);
+
+    List<User> findAllMentors();
 }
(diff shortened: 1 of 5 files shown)


JEspinosa1770 (approved):
The GET request is properly implemented.

With this, we can continue with the tasks.

Perhaps we could add some tests to the filter and the endpoint, but that's up to you.

cristhianchulca49:
@JEspinosa1770 Thanks you for the feedback!. I've add test for endpoint and the filter

mpujazon (approved):
Approved, thanks for the update. The new `GET /mentors` endpoint and mentor-only filtering look good to me.

kenan-it-academy (approved):
I was surprised by the choice of using a `/mentors` URL instead of a payload filter, but I think it's reasonable, actually

PR #1040, opened 2026-06-22
refactor: move /me endpoint from AuthController to UserController

Spook242:
This PR moves the /me endpoint from AuthController to UserController.

Remove the authMe method and its associated logic from AuthController.

Implement the @GetMapping("/me") endpoint in UserController and reuse the mapping logic.

Adjust necessary imports and dependencies.

Updated the Angular authentication service to consume the new endpoint: /api/account/users/me instead of the old route.

All test pass.


cristhianchulca49 (changes requested):
###  Quick Suggestion: Cleaning up duplicate tests

Hey! Nice work restructuring the endpoints to separate authentication from user logic. To keep our test suite clean and green for the demo, let's make two quick adjustments:

* **Remove Duplicate Tests from `UserControllerTest`:** Since the `/api/account/auth/me` endpoint was moved over to the `AuthController`, we should completely delete `shouldReturnUserProfileWhenAuthenticated()` and `shouldReturn401WhenNotAuthenticated()` from the `UserControllerTest` file. They are already being tested perfectly in `AuthControllerTest`, so we don't need them in both places!

Captura de pantalla 2026-06-22 a las 22 38 15

* **Remove the `TestSecurityConfig` block:** The local security configuration added at the bottom of `UserControllerTest` is likely what's causing the test to fail. It overrides our global security settings, which prevents annotations like `@WithMockUser` from working correctly and causes unpredictable behavior. 
Captura de pantalla 2026-06-22 a las 22 39 50


If we delete those duplicate tests and remove that extra configuration block, everything should pass smoothly! Let me know if you want a hand with it.

Spook242:
Thanks for the detailed review, Cristhian...!!! You were totally right. I've removed the duplicate tests and the TestSecurityConfig block. The test suite is clean and passing green globally now.

I needed the tests to pass locally to try something out, and I completely forgot to revert the changes before committing. I'm juggling a few merged branches right now for testing and getting a bit mixed up. Thanks for the review...!!!

isaacmg-bit (approved):

cristhianchulca49 (approved):
Great job! now all the tests pass and there are no duplicate tests.

mpujazon (approved):
Looks good to me, great refactor moving `/me` into `UserController` and updating the Angular client to the new route. Approving.

backend/account-service/src/test/java/com/ita/challenges/account/infrastructure/adapter/web/in/UserControllerTest.java
```diff
@@ -55,31 +48,4 @@ void shouldReturn404WhenUserNotFound() throws Exception {
         mockMvc.perform(get("/api/account/users/unknown/role"))
                 .andExpect(status().isNotFound());
     }
-
-    @Test
```
kenan-it-academy, line 61: Tests should be moved, not removed 😉

kenan-it-academy (changes requested):

backend/account-service/src/test/java/com/ita/challenges/account/infrastructure/adapter/web/in/UserControllerTest.java
kenan-it-academy: Oh wait! My bad, these were duplicate! All good!

kenan-it-academy (approved):

PR #1041, opened 2026-06-22
[TICKET] add mentor field to ticket

cristhianchulca49:
### Description
This PR updates the core domain model to support mentor assignment. It enables the ticket structure to safely store and update the reference to the assigned mentor.

### Technical Changes
* Added an optional `assignedMentorId` (String) field to the `Ticket` aggregate class.
* Updated the `Ticket` constructor, `.create()`, and `.restore()` factory methods to support this new property.
* Defaulted `assignedMentorId` to `null` for all new tickets.
* Ensured data backward compatibility within `.restore()` methods to protect existing tests.
* Added a clean domain method to update/transition the ticket state with the new mentor ID immutably.

PR #1047, opened 2026-06-23
[TICKET] assign a mentor to a ticket

### LINKED ISSUE #1034: [TICKET] Assign a Mentor to a ticket
* **Goal:** Allow a Mentor to officially take responsibility for an unassigned ticket.
* **Acceptance Criteria:**
  * A Mentor can send a request with their username to claim a specific ticket.
  * Security check: Only users with the "Mentor" role can perform this action. If a regular user tries, the system blocks the action (returns a 403 Forbidden error).
  * The ticket successfully updates and saves the assigned Mentor.

cristhianchulca49:
### Description
This PR implements the API endpoint that allows administrators or mentors to officially assign a mentor to an unassigned ticket. It handles payload mapping, authorization checks, and repository updates.

### Technical Changes
* Extended the `TicketController` with a `PATCH /api/account/tickets/{id}/assign`) endpoint.
* Configured the endpoint to accept a payload containing the mentor's identifier (`userName`).
* Integrated the controller action with the repository to persist the ticket's updated state with its new assigned mentor.

⚠️ ALERT  
This PR depends on #1041 because this endpoint requires the domain to be prepared first.

### DIFF
--- a/backend/account-service/src/main/java/com/ita/challenges/account/infrastructure/adapter/web/in/TicketController.java
+++ b/backend/account-service/src/main/java/com/ita/challenges/account/infrastructure/adapter/web/in/TicketController.java
@@ -176,7 +176,7 @@ public class TicketController {
     }
 
     @PatchMapping("/{id}/assign")
-    public ResponseEntity<?> assignMentor(@PathVariable String id,
+    public ResponseEntity<TicketPatchResponse> assignMentor(@PathVariable String id,
                                           @RequestBody TicketAssignRequest assignRequest,
                                           @AuthenticationPrincipal OAuth2User user) {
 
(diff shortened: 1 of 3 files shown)


Urian1983 (commented):
I have to say that I proposed this change a few iterations before, and it was rejected, at last someone saw the light. LGTM!  For me to have a Record/DTO that gives only the needed information is an important upgrade in the security aspect.

Spook242 (changes requested):
Hi Cristhian...!!! Great implementation for the new assign endpoint but there are no test files included.

cristhianchulca49:
@Spook242 You’re right! I had overlooked the tests, but they’re uploaded now.

Spook242:
Awesome...!!! Thanks for adding the tests. The coverage for both the success and error scenarios looks great now. Great job @cristhianchulca49...!!! 👍🏻

Spook242 (approved):

mpujazon (approved):
Approved, this looks good to me. The endpoint, payload mapping, and authorization/repository integration all line up with the intended behavior.

backend/account-service/src/main/java/com/ita/challenges/account/infrastructure/adapter/web/in/TicketController.java
```diff
@@ -172,4 +169,38 @@ public ResponseEntity<TicketPatchResponse> patchTicket(@PathVariable String id,
                 .orElse(ResponseEntity.notFound().build());
     }
 
+    @PatchMapping("/{id}/assign")
```
Urian1983, line 178: - appUser` is retrieved but not used afterward. If the only goal is to validate that the user exists, assigning it to a variable could be avoided.
- I don't see any validation that `mentorAssignedId` actually corresponds to an existing mentor before assigning it.
- It would be interesting to review the permissions: currently, it seems that any authenticated user can assign a mentor to the ticket.
- The validation of `mentorAssignedId` could be delegated to `Bean Validation` (`@NotBlank`) in the DTO to keep the controller cleaner.
- For readability, the flow with `Optional.map()` could be simplified by using `orElseThrow()` and returning the `ResponseEntity` at the end.

Urian1983 (changes requested):

backend/account-service/src/main/java/com/ita/challenges/account/infrastructure/adapter/web/in/TicketController.java
cristhianchulca49: Hi **@Urian1983**! Thanks a lot for the review and your points, they are spot on. I've refactored the endpoint applying most of your feedback:

* **Linear flow with `Optional`:** I replaced `.map(...)` with `.orElseThrow()`. You’re totally right, the flow now reads from top to bottom and is much more readable.
* **Removed `appUser` variable:** We no longer store the instance since we only need to check if the user exists in the database.
* **Mentor validation:** I added a database check to make sure the `mentorAssignedId` actually exists before updating the ticket.

**Regarding the points I couldn't apply as is:**

1. **DTO Validation (`@NotBlank`):** I couldn't move this to the DTO because, for the demo and future features, the admin needs to be able to unassign mentors by sending a `null` or empty value. If we use `@NotBlank`, the framework would reject the request before reaching the controller.
2. **Permissions and Roles:** Since we are preparing this code for the upcoming demo and the front-end doesn't handle or send user roles yet, I preferred to leave the permissions open so it doesn't break the presentation flow. I've added a critical *TODO* to handle security in the next phase.

The changes are pushed! Please take a look whenever you can.

Urian1983 (approved):

kenan-it-academy (approved):

PR #1053, opened 2026-06-24
feat(filter-chip): add reusable filter chip component 1/2 #1009

JEspinosa1770:
## Description
Adds a new reusable `FilterChipComponent` to `shared/components` that displays a toggleable pill-shaped button.

⚠️ This pull request precedes number #1054. Merge this one first before merging the other.

## Changes
- New standalone component with `label` (required) and `active` inputs, and a `clicked` output
- Visual states: default (white background, gray border) and active (dark background, white text)
- `data-testid` attribute for reliable test selection
- Unit tests covering label rendering, default state, active state and click emission

## Out of scope
Integration with parent components and filter logic will be handled in a separate PR.

Issue #1009, opened 2026-06-19
[CHALLENGE] Create language buttons


isaacmg-bit:
We need a button for language filtering:

- [ ] Reusable single button
- [ ] Buttons should be styled according to the Figma design system
- [ ] Clicking an already selected button deselects it

PR #1054, opened 2026-06-24
feat(challenges-list-page): integrate filter-chip component for language filtering 2/2 #1009

JEspinosa1770:
## Description
Second PR of the language filter feature. Integrates the `FilterChipComponent` into `ChallengesListPage` to display the language filter pills.

> ⚠️ This PR depends on #1053. Merge that one first before merging this.
> ⚠️ The checks will be OK when #1053 is merged

## Changes
- Import and declare `FilterChipComponent` in `ChallengesListPage`
- Add `languages` array to iterate the chips in the template
- Add `selectedLanguage` signal to track the active chip
- Add `onLanguageChipClick()` to handle the toggle logic

## Out of scope
Filtering the challenge list based on the selected language will be handled in a follow-up task.


Spook242 (approved):
Nice work on the filtering feature...!!! Just a quick suggestion: could we move the languages array to a separate constant file? What do you think?

JEspinosa1770:
> Nice work on the filtering feature...!!! Just a quick suggestion: could we move the languages array to a separate constant file? What do you think?

Hi Pedro, that's a good idea. I left it where it was because there are already other definitions in the file, but it would be good practice to put them all in a separate file.

mpujazon (approved):
Looks good to me, the filter chip integration is clean and the changes are scoped appropriately. Approved.

isaacmg-bit (approved):
Looks good! We definitely should move all that stuff from the component to their own single files in a future sprint, but you've done well putting it in the component for now.

kenan-it-academy (changes requested):
Missing a test that clicking actually changes the language 😉

JEspinosa1770:
> Missing a test that clicking actually changes the language 😉

Hi, @kenan-it-academy .

In PR #1059, which you've already approved, there's a test for the method implemented in that PR.
The decision to leave the test in that PR stemmed from the fact that the method defined here would undergo a minor modification in the next PR, which would contain the final version. Therefore, I thought testing a temporary method, which was going to be modified soon, was unnecessary.

Let me know if you think it's necessary to include a small test here, even if it's removed in the next PR, and I'll add it.

JEspinosa1770:
> Missing a test that clicking actually changes the language 😉

I've added the test you requested.

ecantosf (approved):
Looks good from a backend-integration perspective. The values in the `languages` array (JAVA, PHP, JAVASCRIPT, TYPESCRIPT, PYTHON, SQL) match exactly the ChallengeLanguage enum on the backend side.

frontend/src/app/features/challenges/pages/challenges-list-page/challenges-list-page.spec.ts
```diff
@@ -107,19 +107,6 @@ describe('ChallengesListPage', () => {
     expect(component.challenges().some(c => c.id === CHALLENGES_MOCK[0].id)).toBe(false);
   });
 
-  it('should set selectedLanguage when a language chip is clicked', () => {
```
kenan-it-academy, line 110: I think you've accidentally removed the tests instead. You're correct that the tests were in another PR so they're unnecessary here I guess? But I disagree with this way of splitting the PRs. It is what it is. Please unremove these tests thought 😉

kenan-it-academy (changes requested):

frontend/src/app/features/challenges/pages/challenges-list-page/challenges-list-page.spec.ts
JEspinosa1770: Okay, I'll add the tests.

kenan-it-academy (approved):

PR #1056, opened 2026-06-25
adding new methods to Submission.java and creating SubmissionBuffer.java

### LINKED ISSUE #1055: [CHALLENGE] Support draft recovery and editing continuity in Submissions
**This is sub-issue is an addendum, is optional, adds a functionality that is needed but will be fullly deployed in a future iteration**

As part of the redesign of the submission flow, an in-memory buffer is 
introduced to hold only the active draft (IN_PROGRESS) for each user, 
separate from the historical repository (SubmissionRepository), which 
will continue to store all submissions (both IN_PROGRESS and SUBMITTED).

This sub-issue covers exclusively the buffer infrastructure piece and 
the changes needed in the Submission domain model to support state 
transitions without breaking its value-object-oriented design. 
Integration with the use case handlers (SaveDraftSubmission, 
FinalizeSubmission, MarkIncomplete) is addressed in a follow-up sub-issue.

Urian1983:
**This is PR is an addendum, is optional, adds a functionality that is needed but will be fullly deployed in a future iteration. This means that is not prioriitary**

## Context
As part of the redesign of the submission flow, an in-memory buffer is 
introduced to hold only the active draft (IN_PROGRESS) for each user, 
separate from the historical repository (SubmissionRepository), which 
will continue to store all submissions (both IN_PROGRESS and SUBMITTED).

This sub-issue covers exclusively the buffer infrastructure piece and 
the changes needed in the Submission domain model to support state 
transitions without breaking its value-object-oriented design. 
Integration with the use case handlers (SaveDraftSubmission, 
FinalizeSubmission, MarkIncomplete) is addressed in a follow-up sub-issue.

### New: `SubmissionBuffer`
- Static utility class (no DI, static `Map<UserId, Submission>`).
- Indexed solely by `UserId` "a user can only have one active draft in 
the entire system, regardless of challenge.
- `save(userId, submission)`: stores the draft; throws 
"If the submission is not in" 
status.
- `findLastByUserId(userId)`: returns the user's active draft, if any.
- `existsByUserId(userId)`: checks whether the user has an active draft.
- `removeByUserId(userId)`: removes the user's draft (e.g., on finalize).
- `clear()`: empties the whole buffer (intended for test use).

### Modified: `Submission`
The following methods are added, with existing ones 
(`createInProgress`, `createSubmitted`) left unchanged:

- `restore(id, challengeId, userId, status, code, createdAt, updatedAt)`: 
  reconstructs a submission, preserving its original status and 
timestamps, intended for deserialization from persistence.
- `toSubmitted(draft)`: creates a new instance from an IN_PROGRESS draft. 
  marking it as SUBMITTED; preserves id/challengeId/userId/code/createdAt 
and refreshes updatedAt.
- `toInProgress(existing)`: creates a new instance from an existing one 
  submission, moving it back to IN_PROGRESS (use case: marking as 
  incomplete); preserves id/challengeId/userId/code/createdAt and 
refreshes updatedAt.
- `updateCode(newCode)`: the class's only mutator; updates `code` and 
``in place. Intended for use only while the submission is 
IN_PROGRESS (invariant enforced by the caller, not by the class itself).

###Notes
Because the limit of 100 lines in every PR the tests for both classes will be add in a future PR

### DIFF
--- a/backend/challenge-service/src/main/java/com/itachallenges/challengeservice/submission/domain/Submission.java
+++ b/backend/challenge-service/src/main/java/com/itachallenges/challengeservice/submission/domain/Submission.java
@@ -45,4 +45,42 @@ public class Submission {
         return submission;
     }
 
+    public static Submission restore(SubmissionId id, ChallengeId challengeId, UserId userId,
+                                     SubmissionStatus status, String code,
+                                     Instant createdAt, Instant updatedAt) {
+        Submission submission = new Submission();
+        submission.id = id;
+        submission.challengeId = challengeId;
+        submission.userId = userId;
+        submission.status = status;
+        submission.code = code;
+        submission.createdAt = createdAt;
+        submission.updatedAt = updatedAt;
+        return submission;
+    }
+
+    public static Submission toSubmitted(Submission draft) {
+        Submission submission = new Submission();
+        submission.id = draft.id;
+        submission.challengeId = draft.challengeId;
+        submission.userId = draft.userId;
+        submission.code = draft.code;
+        submission.status = SubmissionStatus.SUBMITTED;
+        submission.createdAt = draft.createdAt;
+        submission.updatedAt = Instant.now();
+        return submission;
+    }
+
+    public static Submission toInProgress(Submission existing) {
+        Submission submission = new Submission();
+        submission.id = existing.id;
+        submission.challengeId = existing.challengeId;
+        submission.userId = existing.userId;
+        submission.code = existing.code;
+        submission.status = SubmissionStatus.IN_PROGRESS;
+        submission.createdAt = existing.createdAt;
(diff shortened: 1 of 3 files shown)


Spook242 (changes requested):
Great work, Josep...!!! However, while reviewing the SubmissionBuffer implementation, I have an important question regarding the functional architecture.

I see that you've designed the buffer as a utility class with a static ConcurrentHashMap. This could cause issues with testing, and also, if we deploy the app across multiple instances in the future, the static state won't be shared between them.

Wouldn't it be much more robust and functional to turn this Buffer into a @Component or @Repository and inject it via constructor into the use cases that need it? Let me know what you think...!!!

Urian1983:
I came with a better solution to the problem,.

PR #1058, opened 2026-06-25
Implement draft submission lifecycle (save, finalize, mark incomplete)

### LINKED ISSUE #1057: [CHALLENGE] Implement draft submission lifecycle (save, finalize, mark incomplete)
Depends on #1055 

## Context
Sub-issue #1055 introduced `SubmissionBuffer` (the static active-draft buffer) 
and the supporting `Submission` methods (`restore`, `toSubmitted`, 
`toInProgress`, `updateCode`) needed to transition a submission between 
states without breaking its design.

This sub-issue wires that buffer into the three existing use case 
handlers, replacing the placeholder/incomplete logic with the full 
draft → finalize → mark-incomplete flow.

Urian1983:
> **Depends on #1056 ** — requires `SubmissionBuffer` and the 
> `Submission` transition methods (`restore`, `toSubmitted`, 
> `toInProgress`, `updateCode`) introduced there. Do not merge before that PR.

> **No tests included in this PR.** Test coverage for the three handlers 
> is tracked separately / to be added in a follow-up.

## Summary
Implements the use case logic that was left as stubs after the previous 
PR (`SubmissionBuffer` + `Submission` transition methods).

## What changed
- `SaveDraftSubmissionUseCaseHandler`: now checks the buffer for an 
  existing draft before creating a new `Submission`; updates code 
  in-place when a draft is found, otherwise creates one. Persists to 
  both buffer and historical repository on every call.
- `FinalizeSubmissionUseCaseHandler`: fully implemented. Pulls the 
  active draft from the buffer, validates it matches the command's 
  challenge, transitions it to `SUBMITTED`, persists it, and clears 
  the buffer entry.
- `MarkIncompleteUseCaseHandler`: fully implemented. Pulls the 
  submission from the historical repository by id, validates ownership, 
  transitions it back to `IN_PROGRESS`, and writes the result back to 
  both the buffer and the historical repository.

## Why
- Keeps the active-draft buffer as the single source of truth for "what 
  is currently in progress," while the historical repository remains 
  the append-friendly record of everything that happened — avoids 
  mixing two different lifecycles in one data structure.
- Adds two validations beyond the literal ask, both justified on 
  correctness/security grounds rather than style:
  - `userId` ownership check in `MarkIncomplete` — without it, knowing a 
    `submissionId` would be enough to tamper with someone else's 
    submission.
  - `challengeId` consistency check in `Finalize` — cheap to run, turns 
    a possible silent-wrong-challenge bug into an explicit failure.

## Out of scope
- Disk persistence for the historical repository.
- Frontend integration (save/finalize buttons).
- **Test coverage** — explicitly not included in this PR.

### DIFF
--- a/backend/challenge-service/src/main/java/com/itachallenges/challengeservice/submission/application/usecase/FinalizeSubmissionUseCaseHandler.java
+++ b/backend/challenge-service/src/main/java/com/itachallenges/challengeservice/submission/application/usecase/FinalizeSubmissionUseCaseHandler.java
@@ -23,27 +23,18 @@ public class FinalizeSubmissionUseCaseHandler implements FinalizeSubmissionUseCa
 
     @Override
     public void execute(FinalizeSubmissionCommand command) {
-        UserId userId = UserId.of(command.userId());
-        ChallengeId challengeId = ChallengeId.of(command.challengeId());
+        var userId = UserId.of(command.userId());
+        var challengeId = ChallengeId.of(command.challengeId());
 
-        // Buscar el draft activo del usuario en el buffer
-        Optional<Submission> activeDraft = SubmissionBuffer.findLastByUserId(userId);
-        if (activeDraft.isEmpty()) {
-            throw new NoSuchElementException("No active draft found for user: " + userId);
-        }
+        var draft = SubmissionBuffer.findLastByUserId(userId)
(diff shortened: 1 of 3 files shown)


Urian1983:
Out of scope and time

Issue #985, opened 2026-06-18
[CHALLENGE] Filter challenges by difficulty


vicenlu:
<img width="575" height="232" alt="Image" src="https://github.com/user-attachments/assets/6ef783de-3da8-40f7-9c3a-62d065b1fc3f" />

PR #1067, opened 2026-06-26
[CHALLENGE] Implement difficulty filter logic

### LINKED ISSUE #1006: [CHALLENGE] Implement difficulty filter logic
As a user browsing the challenges catalog, I want to filter challenges by difficulty so that I can find ones that match my current skill level.

Acceptance criteria
- When a user selects a difficulty (Easy, Medium, or Hard), the catalog   shows only challenges of that difficulty.
- When no difficulty is selected, the catalog shows all challenges,   preserving the current behaviour.
- When an invalid difficulty value is requested, the system responds   with a clear error rather than returning unfiltered or empty results.

ecantosf:
Closes #1006

Adds support for filtering challenges by difficulty via `GET /api/challenge?difficulty=EASY|MEDIUM|HARD`, combinable with the existing `?language=` filter from #1019.

## Changes
- Replaces `findByLanguage` with `findByCriteria(language, difficulty)` at the repository port; both parameters are optional (null = no filter).
- Adds `?difficulty` query param to `GET /api/challenge`.
- Tests added at repository and controller layers for difficulty filter,  combined filter, and invalid value (400 via Spring enum binding).

## Why a combined method instead of two parallel ones
The UI design (#985, #990) places language and difficulty filters as independent controls that users will naturally combine. A single 
`findByCriteria` keeps the controller free of branching logic and maps cleanly to a future `WHERE` clause when a persistent store is 
introduced.

### DIFF
--- a/backend/challenge-service/src/main/java/com/itachallenges/challengeservice/catalog/domain/port/out/ChallengeRepository.java
+++ b/backend/challenge-service/src/main/java/com/itachallenges/challengeservice/catalog/domain/port/out/ChallengeRepository.java
@@ -3,13 +3,14 @@ package com.itachallenges.challengeservice.catalog.domain.port.out;
 import com.itachallenges.challengeservice.catalog.domain.model.Challenge;
 import com.itachallenges.challengeservice.catalog.domain.valueobject.ChallengeId;
 import com.itachallenges.challengeservice.catalog.domain.valueobject.ChallengeLanguage;
+import com.itachallenges.challengeservice.catalog.domain.valueobject.ChallengeDifficulty;
 
 import java.util.List;
 
 public interface ChallengeRepository {
 
     List<Challenge> findAll();
-    List<Challenge> findByLanguage(ChallengeLanguage language);
+    List<Challenge> findByCriteria(ChallengeLanguage language, ChallengeDifficulty difficulty);
     Challenge update(Challenge challenge);
     Challenge save(Challenge challenge);
     void delete(ChallengeId id);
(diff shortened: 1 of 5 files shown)


Spook242 (changes requested):
Great job on this refactoring Eduard...!!! Moving to a single findByCriteria method is a very clean and scalable architectural decision.

However, I spotted a critical bug in the stream filters. Since ChallengeLanguage and ChallengeDifficulty are Value Objects and using the == operator will not compare their actual values. This means the filter might fail even if the difficulty is exactly the same.

We should use .equals() here to compare the objects safely. What do you think...???

ecantosf:
> Great job on this refactoring Eduard...!!! Moving to a single findByCriteria method is a very clean and scalable architectural decision.
> 
> However, I spotted a critical bug in the stream filters. Since ChallengeLanguage and ChallengeDifficulty are Value Objects and using the == operator will not compare their actual values. This means the filter might fail even if the difficulty is exactly the same.
> 
> We should use .equals() here to compare the objects safely. What do you think...???

Thanks for the careful review! You're absolutely right for class-based Value Objects like ChallengeId — there `==` would compare references and we'd need `.equals()`.

For these two though, ChallengeLanguage and ChallengeDifficulty are Java enums, not class-based VOs (the package name is a bit misleading). Enum constants are JVM singletons, and `Enum.equals()` is declared final with the implementation iterally `return this == other;` — so the two are equivalent by language design. `==` is also null-safe (`null == EASY` returns false vs NPE with `.equals()`), which is why Effective Java recommends it for enums. I followed the same pattern as #1019 for consistency.

If you'd prefer `.equals()` stylistically across the codebase, happy to do it as a separate refactor ticket so we apply it uniformly. Let me know what you think!

cristhianchulca49 (approved):
Great work! 👍 The implementation is clean and well-tested.

Urian1983 (approved):
The change is on the right track: moving from findByLanguage to findByCriteria is consistent with the controller you showed earlier and allows filters to be extended without modifying the controller.

The stream-based implementation is correct as long as the comparison is done using ==, since these are not variables but constants that do not change their memory address and therefore their reference.

kenan-it-academy (approved):

Urian1983 (approved):
...

Spook242 (approved):

kenan-it-academy (approved):

Issue #1011, opened 2026-06-19
[CHALLENGE] Implement language filter logic


isaacmg-bit:
Challenges list should filter by programming language when a language button is selected.

- [ ] When a language button is selected, the challenges list updates to show only challenges matching that language
- [ ] When the same button is clicked again (or "Clear Filters" is triggered), the full list is restored
- [ ] If no language is selected, all challenges are displayed

PR #1070, opened 2026-06-29
feat(challenges-list-page): repair filter challenges by language via backend #1011

JEspinosa1770:
## Description
This PR resolves a merge error where the backend call was dropped. This PR restores the call.


Urian1983 (approved):
Clear and consistent change. Good improvement in readability and behavior. Nothing to add from my side.

isaacmg-bit (approved):

ecantosf (approved):
Confirmed the bug locally: clicking a language chip on `develop` toggles the active state visually but doesn't trigger any GET /api/challenge call — because `onLanguageChipClick` was missing the `loadChallenges(next)` line after the #1054/#1059 merge collision. The `loadChallenges` method itself was still in place, just not invoked from the click handler.

This PR restores the missing call cleanly. Approving.

Spook242 (approved):
Great catch on finding that missing API call after the merge conflict. Great job Jordi...!!!

kenan-it-academy (changes requested):
Does this not require a test? (to ensure the bug does not happen again)

JEspinosa1770:
> Does this not require a test? (to ensure the bug does not happen again)

This code was already validated by you, but it was lost when I merged it. This means that the code already has tests that pass and that are already merged into the develop directory.

This pull request isn't to fix a bug, but rather to add some lines that were lost in the previous merge.

Perhaps I didn't explain it correctly in the pull request description.

In any case, I've added the tests since they also lost a line in the last merge.

kenan-it-academy (approved):

